PR #17, opened 2024-12-31
chore: Update jsonschema usage

Stranger6667:
Hey!

This PR updates some usages of the `jsonschema` crate including the recent `meta` API which is dedicated to meta-schema validation (and is substantially faster than creating a new validator via `Validator::new`). Also, `validation_error_into_owned` is not needed anymore, as `Validator::new` returns `ValidationError<'static>` + there is the `ValidationError::to_owned` method which does exactly what `validation_error_into_owned` was doing.

Another small detail is that I've noticed that the actual validation error is ignored in `JsonColumnMetadata::validate`, so I replaced it with `is_valid` which is usually significantly faster for invalid instances.


ignatz:
Much appreciated, thank you!

PR #35, opened 2025-03-20
Fix: Bump trail to latest and working version (0.8.0)

mdp:
Very minor, but a small issue I ran into when following the README


ignatz:
Thanks!

PR #37, opened 2025-03-21
Fix: Simplify the docker setup

mdp:
Feel free to ignore this suggestion, but I really just wanted to start this up with docker and see it in action, but the docs don't describe the best way to do this. 

I also ran into an error where `docker compose build` looks for ./traildepot in the local path, but can't find it with the relative path pointing to two levels up. I just went ahead and fixed it to build from the example/blog directory, but there may be a better way to do this.

### DIFF
--- a/examples/blog/README.md
+++ b/examples/blog/README.md
@@ -67,6 +67,17 @@ Default users:
 
 ## Instructions
 
+### Get up and running with Docker Compose in under 2 minutes
+
+```bash
+cd $REPO/examples/blog
+docker compose build
+docker compose up
+# Visit https://localhost and ignore the SSL error
+```
+
+### Building each piece of this example project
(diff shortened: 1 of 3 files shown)


ignatz:
Great suggestion and completely addressing the TODO.

The PR had a minor issue by desyncing the workspace lockfile (due to the dep change). I didn't want to waste your time, so I basically took your changes and applied them with attribution. I also took this opportunity to apply a bigger overhaul to the README, remove some obsolete cruft, and fix some issues with the example in a follow-up.

Thanks a lot. If you want to take another look, your Feedback is very much appreciated.

(Tangentially: also added instructions for running TB with [podman](https://github.com/trailbaseio/trailbase/tree/main/deploy/k8s))

PR #57, opened 2025-05-14
chore:Update.gitignore and dependency versions

vootooguru:
• Updated the version of`deno_media_type`in`Cargo.lock`to avoid compilation errors.

• Updated the version numbers of some other crates.

• Modified the`pnpm_run`function.The original function would fail to find the`pnpm`command on Windows.It is necessary to specify the`pnpm`command based on the operating system.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@ node_modules/
 
 # Dev artifacts
 public/
+trailbase-assets/js/admin/pnpm-lock.yaml
+trailbase-assets/js/auth/pnpm-lock.yaml
+trailbase-js/assets/runtime/pnpm-lock.yaml
--- a/trailbase-build/src/lib.rs
+++ b/trailbase-build/src/lib.rs
@@ -56,7 +56,12 @@ fn write_output(mut sink: impl Write, source: &[u8], header: &str) -> Result<()>
 }
 
 pub fn pnpm_run(args: &[&str]) -> Result<std::process::Output> {
-  let cmd = "pnpm";
+  let cmd = if cfg!(windows) {
+    "pnpm.cmd"
+  } else {
+    "pnpm"
+  };
+
   let output = std::process::Command::new(cmd)
     .args(args)
     .output()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1707,9 +1707,9 @@ dependencies = [
 
 [[package]]
 name = "deno_media_type"
-version = "0.2.6"
+version = "0.2.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "480223262efd08f96b3be5f0457c82bac7296e70dc4e7ef7350751f66293812c"
+checksum = "3d9080fcfcea53bcd6eea1916217bd5611c896f3a0db4c001a859722a1258a47"
 dependencies = [
  "data-url",
  "serde",
@@ -3351,9 +3351,9 @@ dependencies = [
 
 [[package]]
 name = "lettre"
-version = "0.11.15"
+version = "0.11.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "759bc2b8eabb6a30b235d6f716f7f36479f4b38cbe65b8747aefee51f89e8437"
+checksum = "87ffd14fa289730e3ad68edefdc31f603d56fe716ec38f2076bb7410e09147c2"
 dependencies = [
  "async-std",
  "async-trait",
@@ -3375,7 +3375,7 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "url",
- "webpki-roots 0.26.11",
+ "webpki-roots 1.0.0",
 ]
 
 [[package]]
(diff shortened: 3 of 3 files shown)


trailbase-build/src/lib.rs
```diff
@@ -56,7 +56,12 @@ fn write_output(mut sink: impl Write, source: &[u8], header: &str) -> Result<()>
 }
 
 pub fn pnpm_run(args: &[&str]) -> Result<std::process::Output> {
-  let cmd = "pnpm";
+  let cmd = if cfg!(windows) {
+    "pnpm.cmd"
```
ignatz, line 60: Pardon my ignorance, doesn't this depend on how you install pnpm? I'm not too familiar with the binary discovery mechanism but the release uses `scoop` and managed to discover `pnpm`, if we change this, could this break for others?

.gitignore
```diff
@@ -13,3 +13,6 @@ node_modules/
 
 # Dev artifacts
 public/
+trailbase-assets/js/admin/pnpm-lock.yaml
```
ignatz, line 16: Having an actual windows user at hand, I'm wondering if there's a way for us to instead do away with:

`trailbase-build/src/lib.rs:109`

In the limited time I tried I didn't find a good way to deal with the concurrent builds resulting in a ERR_PNPM_EBUSY, without just disabling workspace builds.

The current setup is broken, not only because it pollutes the git workspace as you found out. But it also disregards the `<root>/pnpm-lock.yaml`, which results in unpredictable versions.

Maybe you have a better grip on it? Restoring pnpm workspace builds for windows would be awesome.

vootooguru: I don't have sufficient experience with Node.js, but I do have the environment, and I'm happy to help resolve this issue. For users who are just starting to use TB, a smooth start will boost their confidence in using TB. I'll put your suggestion into practice these next couple of days to see if it works. I'm more than willing to help. Practice makes perfect.

trailbase-build/src/lib.rs
vootooguru: > Pardon my ignorance, doesn't this depend on how you install pnpm? I'm not too familiar with the binary discovery mechanism but the release uses `scoop` and managed to discover `pnpm`, if we change this, could this break for others?

Windows users have significantly different habits from users of other systems. Most people are accustomed to downloading and installing various software from the official website (which is the kind of installation effect I have), or using some kind of graphical installer, similar to the App Store. Very few people directly install software using Scoop or Chocolatey. In my circle, I am the only one who would do this for specific issues, and I usually just download the installation package directly.

.gitignore
ignatz: Much appreciated!

For context, I added the special handling only because I was getting ERR_PNPM_EBUSY on the github workflow. Before breaking up the assets it wasn't a problem because all assets were build sequentially in `trailbase_core`. Now both `trailbase-assets` and `trailbase-js` build JS artifacts with pnpm. And since they don't have a dependency edge, they can do it in parallel. I assume this is a PNPM on windows or filesystem particularity.

Ideally, pnpm had a workspace setting to serialize the builds, lock file access, ... or something. Otherwise, as a workaround (since there isn't really a dependency between the crates) we could also restructure the crates or the dependencies.

trailbase-build/src/lib.rs
vootooguru: > 我主要想知道，是否`s/pnpm/pnpm.cmd/`会为其他人打破它，例如github工作流 (和其他包管理器用户)。 在实践中，如果没有一个可靠的约定或发现机制，我们只能做这么多w/o要求用户手动指定他们的pnpm安装？

This is a good idea.Users can specify the value.If they don't specify it,the default value will be used.

.gitignore
vootooguru: > 非常感谢!
> 
> 对于上下文，我添加了特殊处理，只是因为我在github工作流程中获得了ERR_PNPM_EBUSY。在拆分资产之前，这不是问题，因为所有资产都是按顺序构建的`trailbase_core`。现在两者`trailbase-assets`和`trailbase-js`使用pnpm构建JS工件。因为他们没有依赖边缘，他们可以并行。我假设错误是windows或文件系统特殊性的PNPM。
> 
> 理想情况下，pnpm有一个工作区设置来序列化构建，锁定文件访问，...或者别的什么。否则，作为一种解决方法 (因为没有真正的板条箱之间的依赖关系)，我们也可以重组板条箱或依赖关系。

I haven't got to that point yet,because the issue I'm currently dealing with has interrupted my compilation process.I'm not familiar with GitHub workflows,so I'm not sure if I can reproduce the problem that the workflow has shown on my machine.I can only focus on solving the current issue first and then further investigate whether the workflow problem can be reproduced and resolved.

trailbase-build/src/lib.rs
vootooguru: The issue still exists on my machine. I’m not sure if it’s caused by the environment on my machine. What I do now is to modify the command path every time I need to compile tb, specifying it to point to the .cmd file.

Since I’m the only one with this issue here and it only occurs on Windows, I suggest we can simply mention it in the readme. If more Windows users report the same problem in the future, we can consider conditional compilation for different systems to handle it.

What's your opinion?

ignatz: Opinion aside, I wish there was a discovery mechanism. How do other folks deal with the fact that installers and various package manages install executables with different suffixes and in different locations? I would really hope there's a way for to invoke `pnpm` whether the filename is `pnpm` or `pnpm.cmd`. Otherwise, we always fall back to trying both. I'm just not sure what's the best practice on windows. We're certainly not the first ones to run into this.

PR #80, opened 2025-07-10
Fix EmptyHost error on fresh traildepot

eugenefil:


ignatz:
Thanks for looking into this really appreciated :pray: - after some contemplation I decided to push the fallback responsibility further down the stack to the consumers (email, oauth, ...) rather than providing a generic fallback in more cases.
This allows the consumers to fall-back, warn or fail in more meaningful ways: https://github.com/trailbaseio/trailbase/commit/103ee9d5bdbb4470ba5516f0e90e998fd33e795a

Sorry for the inconvenience. Hope this approach makes sense, otherwise please step on my toes. Much appreciated :pray:

eugenefil:
Fixed by a84c5784b17b5a9d8cfef3425e5a216edd3a9587

PR #91, opened 2025-07-13
Fix: put Send Email Verification button inside form

eugenefil:
When changing email, Send Email Verification button does not do anything b/c it's outside the form. Put it inside.


ignatz:
Much appreciated!

PR #92, opened 2025-07-14
Change the order so that alias of `trail` works properly.

imjlk:
Correct a minor typo? Currently, running the trail alias repeatedly executes mkdir traildepot, which prevents the command from working properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,8 +103,8 @@ curl -sSL https://raw.githubusercontent.com/trailbaseio/trailbase/main/install.s
 Alternatively, run TrailBase using the Docker image:
 
 ```sh
+mkdir traildepot &&
 alias trail='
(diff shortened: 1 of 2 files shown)


imjlk:
Aha, yes can use the `mkdir -p` option as well. But in this case, wouldn’t it be possible to keep creating new traildepot from any directory using the trail command as long as the shell session is active? That might be a bit embarrassing.

ignatz:
> But in this case, wouldn’t it be possible to keep creating new traildepot from any directory using the trail command as long as the shell session is active?

Yes, as long as the alias exists you will create new `./tralidepot`s from anywhere. IMHO, that's intended and the behavior that you would see if you installed `trail` into `$PATH` rather than using docker. Would you have expected a different behavior?

PR #94, opened 2025-07-14
docs: reorganize sidebar navigation

ixxie:
As a new user, I found a lot of useful resources in the docs; however, I didn't always find it easily. This reorganizes the docs in an effort to improve their navigability.

Some pages are combined, moved and renamed but no content is removed.


ixxie:
@ignatz I would say just give me all the feedback in one go! I'm happy to do this quickly over the next day or two.

I agree about `getting started`; I ended up doing it this way, because I saw the need to group several other pages under `why trailbase` and this left the `getting started` section with only one page.

I now pushed another version: tell me if you like it and what changes you require...

ixxie:
> I'm struggling a bit to efficiently give feedback given the breadth - I really just hope I'm not draining your energy - let me try...

Don't worry, I can handle it :smile: 

>     1. You probably folded the "JS API" doc because it was very daft and crappy. No argument here. I was hoping to expand them over time. At the end of the day, alongside record APIs they are the main way of adding endpoints to TrailBase, so I thought it would make sense to have them on the same level 🤷‍♀️ . Eventually, the JS documentation will probably outgrow the record APIs as they are a lot more flexible. (As the framework matures there may be a third dedicated entry for the Rust APIs... speculation, only time will tell) .... long story short, we can go along with your proposal and pull it back out if and when it makes sense 🤷‍♀️

My main motivation here was to improve navigability. I believe this confusing navigation is the main obstacle to understanding Trailbase. I agree eventually the JS API could be expanded, but having one page where all extension options - including the JS API - are compared is useful imo. So I would say, if you at some point want to go into details, leave the JS section on that page and add a new page with more details.

>     2. This feels similar, you probably folded them because especially the SupaBase comparison is very daft and crappy. I see this more as a TBD as an opportunity. It's not clear to me what the benefit is of merging the two. There may be more in the future (FWIW I did look into AppWrite, but I couldn't get AppWrite to work properly... talking about bleeding edge 😅 ) ... long story short, I would lean leaving it as is.

I feel it is better to reduce the number of pages in order to make it easy to navigate. Its also nice to see both together, because if you are shopping around for a backend you may be considering all three (as I have). 

>     3. I think making the goals doc more front and center could make sense. I'm also a bit self-contentious because I really don't like the word-salad I produced in this doc. I'm torn, especially in its current state i don't hate having the install more of an entry point.

I think the solution I settled on is quite nice: _Getting started_ as the first section and _Why TrailBase?_ as the last. The last item on a menu is - IMO - the second most visible after the first. So I think putting it in the end is the right choice.

>     4. I don't understand the grouping. **I understand and like your intent of answering the question "why should I care about using TrailBase" earlier and more prominently**. I think that keeping that separate from noise such as faq, roadmap, ...  may even help stand out more. My approach would have been to rework or replace the "goals" doc to specifically answer "why" and make that more prominent, maybe first, in the getting started flow. My assumption is that "getting started" should lead any prospect user through the essential information in the order we'd like them to. In the proposed change it would be separate from getting started.

I think what all of these pages have in common is that _they inform a prospective user whether and why Trailbase is the right tool for them_. Perhaps we can think of a better title, but I think its a good idea to try and simply the navigation, and a big part of that is trying to reduce the number of sections and pages where possible and avoiding nested menus.

With all these arguments in mind, I propose that you make a final call about the navigation structure (including sections and page titles) and write it in a response. Then I will adjust the PR to follow your instructions.

ixxie:
I see where your coming from on all this. I think we should close this PR, and you should open your own at some point to address these issues the way you see fit. If you would like, I could have a look at that PR and then point out the pain-points from my perspective.

PR #103, opened 2025-07-26
Fix: OAuth2 login via GitLab

eugenefil:
I checked OAuth2 with GitLab. It used unknown to GitLab `identify` scope and expected slightly different format of user info. It seems GitLab changed things since the time the code was originally written.

I had to use `read_user` scope and `state` attribute instead of `active`.


ignatz:
Much appreciated :pray:

PR #108, opened 2025-07-27
Fix profile link

eugenefil:
Fix redirect to profile in /verify_email/confirm endpoint.


ignatz:
Thanks! :pray: Sometimes I feel silly. This isn't the first time. Short of testing, I should just replace all these string constants with a central repository.

Issue #124, opened 2025-08-06
Feature request: add batch & transactional (create/update/upsert/delete)


ibilux:
Hello,

Would it be possible to add support for batch and transactional record operations (create, update, upsert, delete) in a single request, similar to how [PocketBase handles it](https://pocketbase.io/docs/api-records/#batch-createupdateupsertdelete-records)?

PocketBase allows batch operations via both `application/json` or `multipart/form-data`.

A possible example for Trailbase:

```json
{
  "requests": [
    {
      "method": "POST",
      "url": "/api/records/v1/{name}",
      "headers": {},
      "body": { "title": "test1" }
    },
    {
      "method": "PATCH",
      "url": "/api/records/v1/{name}/{record}",
      "headers": {},
      "body": { "title": "updated title" }
    },
    {
      "method": "DELETE",
      "url": "/api/records/v1/{name}/{record}"
    }
  ]
}
```

The existing logic in `create_record_handler`, `update_record_handler`, and `delete_record_handler` could be reused or refactored to handle each request in the batch (whether in the current Axum routes or on a new endpoint).

As always, Thank you for your amazing work on Trailbase.

ignatz:
Different APIs as well... still organizing my thoughts. Do you have a pointer to PockBase's documentation? Only found mentions of hooks and custom handlers.

One thing to note is that any such API would be more limiting than a custom handler for the lack of conditional behavior.

Just curious, what use-cases do you have in mind? Appreciated

ignatz:
>  Do you have a pointer to PockBase's documentation? Only found mentions of hooks and custom handlers

Sorry, searching for "batch" rather than "transaction" did the trick:

https://pocketbase.io/docs/api-records/#batch-createupdateupsertdelete-records

ignatz:
We now have an abstraction to execute a batch of arbitrary mutations. There's probably still some work to be done around access checking. At first I thought, it would make sense to check permissions for all mutations upfront, but that's probably not what one would want. For example, if you added yourself to some permissions group to then to a follow-up mutation, the access check should probably be delayed and when it fails the transaction be rolled back

ignatz:
Did a PoC. It's not too terrible. 

The originally proposed APIs used HTTP method, URLs and headers. That felt a bit like a leaky implementation detail, i.e. an HTTP API depend on HTTP APIs. My PoC might be a bit more idiomatic 🤷‍♀️ , with the only lations over the originally proposed:

[...]

PR #129, opened 2025-08-11
Add multipart-form support to transactions

ibilux:
Hello @ignatz,

This aim to add multipart-form support to transactions #124, it's based on a similar Pocketbase implementation.
The Multipart file support is using indexed field names `files.{operationIndex}.{columnName}`.

The `processFilesFromOperations()` method loops through all operations and scans operations for `File` objects and generates indexed field names using the pattern `files.{operationIndex}.{columnName}` and replaces `File` objects with metadata objects for JSON serialization.

The endpoint then parses the operations from the multipart field name and group files by operation index and then pass the operation-specific files to each `LazyParams::new()` call.

Example with automatic file handling:
```ts
const client = new TrailBaseClient(/* config */);  
const transaction = client.transaction();  
  
// Files are automatically detected and handled  
const avatarFile = new File([...], 'avatar.jpg', { type: 'image/jpeg' });  
const bannerFile = new File([...], 'banner.png', { type: 'image/png' });  
  
transaction.api('users').create({   
  name: "John Doe",   
  email: "john@example.com",  
  avatar: avatarFile  // File automatically detected and indexed as "files.0.avatar"  
});  
  
transaction.api('profiles').update('abc123', {   
  bio: "Updated bio",  
  banner: bannerFile  // File automatically detected and indexed as "files.1.banner"  
});  
  
transaction.api('posts').delete('xyz789');  
  
const result = await transaction.send();
```

It feels like a leaky implementation like you mentioned before, but if it works for Pocketbase i don't see why it won't work for Trailbase (until we find a better solution i guess).

**Note:** This is just a draft, further testing and checks should be done.

### DIFF
--- a/crates/assets/js/client/src/index.ts
+++ b/crates/assets/js/client/src/index.ts
@@ -428,6 +428,9 @@ export interface Client {
   ///
   /// Unlike native fetch, will throw in case !response.ok.
   fetch(path: string, init?: FetchOptions): Promise<Response>;
+
+  /// Creates a new transaction batch.
+  transaction(): TransactionBatch;
 }
 
 /// Client for interacting with TrailBase auth and record APIs.
@@ -483,6 +486,11 @@ class ClientImpl implements Client {
   /// Construct accessor for Record API with given name.
   public records<T = Record<string, unknown>>(name: string): RecordApi<T> {
     return new RecordApiImpl<T>(this, name);
+  }
+
+  /// Creates a new transaction batch.
+  public transaction(): TransactionBatch {
+    return new TransactionBatch(this);
   }
 
   public avatarUrl(userId?: string): string | undefined {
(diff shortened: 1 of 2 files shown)


ibilux:
> Which part do you consider leaky?

The part of using indexed field names `files.{operationIndex}.{columnName}` to add multipart file support. But, as I said, Pocketbase is using the same idea for a long time now and it's working with no problem and no complains.

> I'm wondering how you'd feel about embedding the file contents into JSON, rather than the other way round? Are you looking at it from an efficiency point of view?

Tbh, I was amazed that Trailbase already handles files by embedding them into JSON -kudos for you for thinking ahead-, and for small files or blobs this will be great. But for large files we should consider the ~33% Base64 encoding increases in file data, and also, the cpu and memory usage for encoding the data.

I love how the project benchmark shows how powerful it is, and I think we should do whatever we can to keep things efficient.

> will affect a few more client libraries.

I thought about that and I already looked to the other clients to see if I can add this to them, but unfortunately i don't have much expertise in the other clients programming languages.

ignatz:
Based on your response, I'll carefully assume that you primarily wanted file support but also care about efficiency.

> But, as I said, Pocketbase is using the same idea for a long time now and it's working with no problem and no complains.

I'm sure it works. But I certainly can offer concerns. If we're jumping straight to "cloning PocketBase", at best we're lazy and at worst we're missing out on a lovely discussion :). I'd also like to stay away from authority arguments.

A few thoughts:

 * Supporting multipart-form for single record mutations makes sense to support simple web forms, e.g. insert results of a survey via plain HTML. For batches, that's not the case.
 * Since, we're not catering to plain HTML, multipart-forms are a pretty archaic format intimately tied to early browsers and mime.
 * PocketBase is very web-centric, TrailBase lesser so. If something is not running in the browser, there's much less incentive to fall back to MIME-tech for binary encoding (I understand they are convenient in the browser).
 * Also their streaming aspect is irrelevant for this use-case.
 * While I do care about efficiency, both for payload and protocol, I care about a lot of things :)
 * I care about type-safety

For Browsers, JSON is convenient, too. As you correctly point out, it stinks for binary encoding. Doesn't mean there aren't other options than base64, e.g msgpack, bson, protobuf, ... . Clearly there are other costs like increased dependency footprint for the client, ... 

I'm mostly just trying to understand how important this is, for what, and based on what, what would make for the best trade-offs. Just for context, my head, the biggest blocker for rolling out an early version of the transaction feature was to support the clients.

Tangent: If we wanted to make file uploads more efficient we should probably support presigned URLs :hide:

ibilux:
> I'm sure it works. But I certainly can offer concerns. If we're jumping straight to "cloning PocketBase", at best we're lazy and at worst we're missing out on a lovely discussion :). I'd also like to stay away from authority arguments.

That’s exactly why I opened this as a draft, because I love constructive and thoughtful conversation.

I had other ideas for efficient file handling, such as:

1. Extending the Operation enum to include optional operation IDs
2. Using a nested JSON structure with file references
3. Using separate multipart sections per operation4. 

But all of these still rely on multipart for sending binary data. For comparison, the existing Record API endpoints already support multipart through the Either extractor. From a consistency perspective, it seemed natural to make the transactions endpoint support multipart in the same way.

> * Supporting multipart-form for single record mutations makes sense to support simple web forms, e.g. insert results of a survey via plain HTML. For batches, that's not the case.
> * Since, we're not catering to plain HTML, multipart-forms are a pretty archaic format intimately tied to early browsers and mime.
> * PocketBase is very web-centric, TrailBase lesser so. If something is not running in the browser, there's much less incentive to fall back to MIME-tech for binary encoding (I understand they are convenient in the browser).
> * Also their streaming aspect is irrelevant for this use-case.
> * While I do care about efficiency, both for payload and protocol, I care about a lot of things :)

True, I was looking at the problem from a web-dev point of view, and transactions are indeed less tied to plain HTML forms than single record APIs.

> Just for context, in my mind, the biggest blocker for rolling out an early version of the transaction feature was to support the clients (in a type-safe manner).

I agree, most languages don’t handle multipart combined with typed operations very cleanly. That’s actually why I had some trouble trying to port this to other client libraries.

In conclusion, I think we shouldn’t over-engineer this early. File handling via JSON embedding already works and should be more than enough for now. We can revisit multipart, or an alternative binary-friendly format, only if real-world usage shows that payload size or performance becomes a bottleneck.

ignatz:
> > I'm sure it works. But I certainly can offer concerns. If we're jumping straight to "cloning PocketBase", at best we're lazy and at worst we're missing out on a lovely discussion :). I'd also like to stay away from authority arguments.
> 
> That’s exactly why I opened this as a draft, because I love constructive and thoughtful conversation.

Much appreciated :pray: 
 
> I had other ideas for efficient file handling, such as:
> 
>     1. Extending the Operation enum to include optional operation IDs
> 
>     2. Using a nested JSON structure with file references
> 
>     3. Using separate multipart sections per operation4.

Do you have a preference, in case we end up going with multipart now or eventually?

> But all of these still rely on multipart for sending binary data. For comparison, the existing Record API endpoints already support multipart through the Either extractor. From a consistency perspective, it seemed natural to make the transactions endpoint support multipart in the same way.

I can see where you're coming from. It certainly feels like the natural choice from that point of view.

At the same point, the clients currently don't even support multi-part file-uploads for single operations. Maybe an oversight on my end. Either way, if we wanted to tackle this both for single and batch operations, it would be very weird, and certainly not consistent, if clients would send JSON when no file is attached and multipart otherwise :sweat_smile: 
 
> In conclusion, I think we shouldn’t over-engineer this early. File handling via JSON embedding already works and should be more than enough for now. We can revisit multipart, or an alternative binary-friendly format, only if real-world usage shows that payload size or performance becomes a bottleneck.

FWIW, I'm happy with either approach: tackling it in time or tackling in foresight, both valid.

I guess the one realization I had just now, and thanks for chatting with me, is that if we want to tackle more efficient file uploads we should probably first think about what we want to do for the single-op endpoints. Always json, always multipart, hybrid, something else entirely? :pray:

ibilux:
> Do you have a preference, in case we end up going with multipart now or eventually?

Personally, I’m leaning toward the indexed field naming approach `files.{operationIndex}.{columnName}` for multipart. It solves the collision problem cleanly and works nicely with what we already have. And `LazyParams::new()` can already handle this without any signature changes.
Another plus: the server logic can still fall back to the current JSON-embedded files, so we don’t lose compatibility for existing clients.

> At the same point, the clients currently don't even support multi-part file-uploads for single operations. Maybe an oversight on my end. Either way, if we wanted to tackle this both for single and batch operations, it would be very weird, and certainly not consistent, if clients would send JSON when no file is attached and multipart otherwise 😅

I didn’t realize until now that some clients don’t even support multipart for single-operation endpoints, I just assumed they did because the server already handles it. That’s actually an important piece in deciding how to move forward.

> I guess the one realization I had just now, and thanks for chatting with me, is that if we want to tackle more efficient file uploads we should probably first think about what we want to do for the single-op endpoints. Always json, always multipart, hybrid, something else entirely? 🙏

Right now, TrailBase clients file handling uses Base64. It’s simple, works everywhere, and dependency-free. But, I haven’t seen explicit file size limits in the codebase, hence, for large files this will comes with trade-offs:

1. ~33% size overhead
2. Higher memory use (files must be loaded entirely into memory for processing, which can be problematic for large files)
3. Extra CPU Overhead for encoding/decoding
4. Slower JSON parsing for large blobs
5. No streaming for big files

I think multipart is the strongest alternative. It may be an older standard, but that’s a good thing, it’s proven, reliable, and **built into HTTP itself**. Since Trailbase endpoint APIs already support multipart requests, it would fit naturally into the existing design without major changes. Benefits include:

1. Universal HTTP support: Works in virtually every HTTP client library.
2. Language agnostic: Consistent across Go, Rust, .NET, Dart, etc.
3. More efficient: No Base64 overhead.
4. Streaming support: Can handle large files without loading them fully into memory if we ever need it.

Type-safety would remain a concern, but it could be mitigated by well-designed client libraries.

That’s why I think a **Hybrid Approach** would be the most practical universal solution and give us the best of both worlds. For small files we can use Base64 in JSON and for large files we can use Multipart form data. This balances compatibility, simplicity, and performance.
We can add a configurable threshold to determine “small vs. large”, or just use a boolean flag in the client library so the user decides whether to embed or use multipart. That way, we’re not forcing one approach on everyone.

PR #132, opened 2025-08-21
Fix a update operation `lazy_params`

ibilux:


ignatz:
I swear, sometimes I can read :hide:.... the formatter. Will just fix it

PR #158, opened 2025-09-28
Allow multiple filters on the same column

kokroo:
## Summary
This PR fixes a bug where applying multiple filters to the same column (e.g., `$gte` and `$lte` for date ranges) would fail. The filter parsing logic now properly handles multiple operators on the same column by converting them into an AND composite filter.

## Problem
Previously, API calls with date range filters would fail:
```
filter[contact_datetime][$gte]=2025-09-25&filter[contact_datetime][$lte]=2025-09-27
```

The issue was that `serde_value_to_single_column_rel_value` in `filter.rs` only handled maps with exactly one operator per column.

## Solution
Modified the filter parsing logic in two locations within `filter.rs` to:
1. Detect when multiple operators are applied to the same column
2. Automatically convert them into an AND composite filter
3. Process each operator individually and combine the results

### Code Changes
- **`crates/qs/src/filter.rs`**: Added logic to handle multiple operators on the same column
- **`crates/qs/src/query.rs`**: Added test case to verify date range filtering works correctly

### Documentation Updates
Updated all client library examples to demonstrate filtering on the same column multiple times:
- **curl**: URL parameters with multiple filters
- **TypeScript**: Array of filter objects  
- **Python**: Filter objects with CompareOp
- **Rust**: `with_filters` with multiple Filter instances
- **Go**: FilterColumn structs
- **Dart**: Filter objects
- **Swift**: Filter enums
- **.NET**: Filter objects with CompareOp

Each example now shows:
- Multiple filters on `watch_time` column (between 90 and 120 minutes)
- Multiple filters on `release_date` column (date range 2020-2023)

## Testing
✅ Successfully tested with a running Trailbase server:
- Created test tables with datetime columns
- Verified single date filters continue to work
- Confirmed date range filters now work correctly
- Tested with various timestamp formats (date-only and full timestamps)

## Test Examples
```bash
# Date range query that now works
curl 'http://localhost:4000/api/records/v1/lucid_panda?filter[contact_datetime][$gte]=2025-09-25&filter[contact_datetime][$lte]=2025-09-27'

# Returns records within the specified date range
```

## Impact
This fix enables common use cases like:
- Filtering records within a date range
- Applying multiple numeric constraints (e.g., price between min and max)
- Any scenario requiring multiple operators on the same column

The change is backward compatible - single operator filters continue to work as before.

This fix was sponsored on company time, thanks to  [Workflow Architects ](https://www.workflowarchitects.io):)

### DIFF
--- a/crates/qs/src/filter.rs
+++ b/crates/qs/src/filter.rs
@@ -109,9 +109,27 @@ where
               Value::Map(BTreeMap::from([(Value::String(key), v)])),
               depth + 1,
             ),
-            _ => Ok(ValueOrComposite::Value(
-              serde_value_to_single_column_rel_value::<D>(key, v)?,
-            )),
+            _ => {
+              // Handle case where v might be a map with multiple operators
+              match v {
+                Value::Map(ref m) if m.len() > 1 => {
+                  // Multiple operators on same column (e.g., $gte and $lte)
+                  // Convert to AND composite of multiple conditions
+                  let conditions = m
+                    .iter()
+                    .map(|(op_key, op_value)| {
+                      let single_op_map = BTreeMap::from([(op_key.clone(), op_value.clone())]);
+                      serde_value_to_single_column_rel_value::<D>(key.clone(), Value::Map(single_op_map))
+                        .map(ValueOrComposite::Value)
+                    })
+                    .collect::<Result<Vec<_>, _>>()?;
+                  Ok(ValueOrComposite::Composite(Combiner::And, conditions))
+                }
+                _ => Ok(ValueOrComposite::Value(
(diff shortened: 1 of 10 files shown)


ignatz:
Awesome find! Thanks so much @kokroo - sorry there wasn't test coverage already... big blunder

I took your patches, slightly modified them, and rebased them in https://github.com/trailbaseio/trailbase/commit/02915446e52c4146babb7dbfd9739fd214dcc33b :pray: 

As for the changes: 
 * I flattened the special casing in the parsing a little
 * I kept the changes out of the per-language examples. Mostly would like to keep them simple. I hope it's clear that multi-filtering should be allowed, as it was to you, minus the egregious bug you hat to fix :pray: 
 
Much appreciated

PR #166, opened 2025-10-14
Add Base64 encoding/decoding functions to SQLite extensions

ibilux:
Hello @ignatz,

This PR introduces two new SQLite extension functions for base64 encoding/decoding of BLOB values:

- `b64_text(blob)` - Converts BLOB to URL-safe base64 string
- `b64_parse(text)` - Converts URL-safe base64 string to BLOB

These functions are similar in to `uuid_text` and `uuid_parse`, but are intended to assist with debugging and inspecting records in the SQL editor where IDs are represented as Base64 strings.

Best regards,

### DIFF
--- a/crates/extension/src/b64.rs
+++ b/crates/extension/src/b64.rs
@@ -1,8 +1,8 @@
 use base64::prelude::*;
 use rusqlite::Error;
-use rusqlite::functions::Context;
-use rusqlite::types::{Type, Value, ValueRef};
 use rusqlite::Result;
+use rusqlite::functions::Context;
+use rusqlite::types::{Value, ValueRef};
 
 /// A URL-safe base64 similar to the SQLite built-in `base64()` extension.
 /// - BLOB → TEXT (encode)
@@ -52,7 +52,7 @@ pub(super) fn base64_url_safe(context: &Context) -> Result<Value> {
 #[cfg(test)]
 mod tests {
   use base64::prelude::*;
-  use rusqlite::{Connection, Error};
+  use rusqlite::Error;
   use uuid::Uuid;
 
   #[test]


ignatz:
> Hey @ignatz 👋 great to hear from you too!
> 
> > * We should probably add some notion of "url safe" to the name, otherwise this will likely become a gotcha when folks expect RFC compliant encoding.
> 
> You're right about the naming. The “url safe” part should definitely be explicit to avoid confusion with standard Base64. I had the same concern, which is why I initially went with the shorter `b64` naming trying to reduce confusion. But, something like `base64_url_safe_encode` / `base64_url_safe_decode` would definitely be clearer.
> 
> > * I appreciate that you aimed at being consistent with UUID 👍 . That said, I'm not sure the verbs quite fit here, i.e. if you have a valid UUID you can either text or byte-encode it - it's a conversion of sorts. Whereas, any blob can be base64 encoded, that doesn't make a blob base64. However, you can decode a certain strings back to a blob... sorry for the sticklering, I'm sure you know all this.
> 
> I wouldn’t call it “sticklering”, I genuinely enjoy and appreciate this kind of detail and deep dive. These discussions and new perspectives are always valuable, make the project better, and are a great learning opportunity.
> 
> When I was writing this PR, I was also thinking about how this could be extended for other scenarios, but I didn’t want to deviate too much from the main goal. Still, generalizing it to handle encoding and decoding more broadly would definitely be the better long-term approach.
> 
> > * Note that SQLite itself has a "base64()" optional extension: https://sqlite.org/base64.html. They have one function and depending on the input data-type the de- or encode. Not sure how I feel about it, just wanted to mention.
> > 
> > TL;DR: maybe `base64_url_safe_(decode|encode)` or if we wanted to be in-line with SQLite's own: `base64_url_safe`, in which case, if we added a non-url-safe version folks naively searching the webs would just get a working `base64` 🤷‍♀️
> > if we wanted to be in-line with SQLite's own: `base64_url_safe`
> 
> Mainly, I was aiming for something lightweight and always available within our existing extension build, primarily for inspection/debugging purposes. But I agree, using a single `base64_url_safe` function that aligns with SQLite’s optional `base64()` extension makes sense. It would also leave room for a future full `base64()` integration if we ever want that.
> 
> > Lastly but maybe most importantly, I'd like to understand this better. I'm not sure how you'd use these functions for this use-case, but we could easily display blobs/ids in various encodings in the UI (w/o invoking sqlite). What would be the ideal state for you? (The UI could use a lot of brush-up)...
> 
> The `id` columns are stored as BLOBs but represented as URL-safe Base64 strings (for UUIDv7-based IDs) in the client libraries. When manually querying or debugging data, it’s handy to be able to do quick conversions in the SQL editor like:
> 
> ```sql
> SELECT * FROM table WHERE id = base64_url_safe('AZlIy6iKcnGmDGYDy-538w==');
> ```

This is a great use-case - thanks!

> I agree that the UI could eventually handle this automatically, but having the function available in SQL makes debugging much smoother. It allows verifying that certain IDs match what’s shown in the app without needing to search through the UI.

Your point makes me wonder if we should a hex-encoded representation instead, e.g. `X'0001'`, then the conversion wouldn't be needed. Otherwise users would need to know about `base64_url_safe`. We should then probably also allow entering blobs in a hex-encoded fashion.
 
> I also think having `Base64` helpers could be handy in custom endpoints too, for encoding or decoding IDs and payloads directly in SQL without needing extra backend code.

For these cases, however, I would never use SQLite. You should be able to use your favorite lib and compile everything to WASM, no need to call the host for such a simple, hermetic operation.

Coming back to the extension, so `base64_url_safe`? Sounds good to me. I still think it would good for users not to have to know about it.

I'd be happy to look into the hex encoding. WDYT?

ignatz:
> Base64 just happens to be more aligned with what client libraries and external APIs use. But, if hex can simplify the default experience, I’m all for it.

I do wonder how much that consistency matters. As long as you're only using a client library it's opaque. It only shows when crafting your own requests. Base64 does have a better compression ratio but we could certainly allow hex-encoding on:

* request bodies on writes,
* list filter parameters,
* and have a query parameter on reads.

The worse compression ratio may exacerbate url length limits for filter queries :woman_shrugging: (...not sure how many folks even filter on blobs).

**I guess the biggest issue would be similar to your use-case, someone wants to copy&paste an id from the admin UI into a request that they're crafting...**

...thanks for pointing this out :pray: - would have missed entirely... genuinely not sure what's best. If I would start from scratch, I would probably require blobs to be JSON objects rather than strings, .e.g

```json
{ "base64": "..." }
```

and 


```json
{ "hex": "..." }
```

However now, this would be a pretty breaking change affecting reads, writes, all the clients and potentially users' own query builders/abstractions. I will need to ponder a little longer :thinking: 

> In that case, we might not even need base64_url_safe at all.

You let me know. This is pretty self-contained, so I'd be happy for us to add `base64` and `base64_url_safe`

ibilux:
> I do wonder how much that consistency matters. As long as you're only using a client library it's opaque. It only shows when crafting your own requests. Base64 does have a better compression ratio but we could certainly allow hex-encoding on:
> 
> * request bodies on writes,
> * list filter parameters,
> * and have a query parameter on reads.
> 
> The worse compression ratio may exacerbate url length limits for filter queries 🤷‍♀️ (...not sure how many folks even filter on blobs).

> However now, this would be a pretty breaking change affecting reads, writes, all the clients and potentially users' own query builders/abstractions. I will need to ponder a little longer 🤔

I agree, the consistency argument mainly matters when someone interacts with the API directly rather than through a client SDK. The `{ "base64": "..." }` / `{ "hex": "..." }` approach would’ve been better if it were there from the start, but at this stage it would be too disruptive and a significant breaking change. The current `{ "blob": "..." }` structure already covers most (if not all) use cases well.

> You let me know. This is pretty self-contained, so I'd be happy for us to add `base64` and `base64_url_safe`

I think adding `base64_url_safe` (and `base64` if needed) is a solid middle ground for now. It gives flexibility without breaking anything, and we can still revisit a more native hex representation later if it proves useful.

On a related note, one case I had in mind (while sleeping on it 😅) was embedding another object inside a record in a custom endpoint using custom SQL queries directly, for example:

```sql
  SELECT
    products.*,
    JSON_OBJECT(
      'id', base64_url_safe(users.id),
      'name', users.name
    ) AS user
  FROM products
  LEFT JOIN products ON products.user = users.id
  WHERE products.id = id;
```

I previously handled this kind of transformation on the JS side, but doing it directly in SQL feels much cleaner and more direct (unless there’s a better approach you’d suggest?). 

Ultimately, I think `base64_url_safe` would be very useful because client libraries already use URL-safe Base64 for binary IDs. Having this function available would improve interoperability between Trailbase and client APIs while keeping the SQL workflow smooth.

Out of curiosity, when you mentioned earlier that _**it would be good for users not to have to know about `base64_url_safe`**_, did you mean that ideally the client/UI would handle the encoding transparently so users wouldn’t need to call it manually? Or were you thinking it should stay hidden to avoid potential misuse or confusion?

ignatz:
> On a related note, one case I had in mind (while sleeping on it 😅) was embedding another object inside a record in a custom endpoint using custom SQL queries directly, for example:
> 
> ```sql
>   SELECT
>     products.*,
>     JSON_OBJECT(
>       'id', base64_url_safe(users.id),
>       'name', users.name
>     ) AS user
>   FROM products
>   LEFT JOIN products ON products.user = users.id
>   WHERE products.id = id;
> ```
> 
> I previously handled this kind of transformation on the JS side, but doing it directly in SQL feels much cleaner and more direct (unless there’s a better approach you’d suggest?).

I think this is another excellent use-case, both for custom endpoints but also `VIEW`-based APIs. Depending on how much more your endpoint does, a `VIEW` might even work for you.

> Ultimately, I think `base64_url_safe` would be very useful because client libraries already use URL-safe Base64 for binary IDs. Having this function available would improve interoperability between Trailbase and client APIs while keeping the SQL workflow smooth.

:+1: 

> Out of curiosity, when you mentioned earlier that _**it would be good for users not to have to know about `base64_url_safe`**_, did you mean that ideally the client/UI would handle the encoding transparently so users wouldn’t need to call it manually? Or were you thinking it should stay hidden to avoid potential misuse or confusion?

Only the former. It's not just about having to call it, but your initial use-case is locked behind users having to known that such an extension function exists.

ibilux:
Hello @ignatz,
I have refined the URL-safe base64 into one function `base64_url_safe()` implementation to match SQLite `base64()`. WDYT?
I see you are working on brushing-up the UI in the [typed_admin_api](https://github.com/trailbaseio/trailbase/tree/typed_admin_api) branch. Did you test the hex-encoded representation?

ignatz:
Sorry, I've been sitting on my hands. I'll review your change first thing tomorrow.

> I see you are working on brushing-up the UI in the [typed_admin_api](https://github.com/trailbaseio/trailbase/tree/typed_admin_api) branch. Did you test the hex-encoded representation?

Kind of. I wanted more typed APIs for a while. So, I took this discussion here as an excuse, i.e. how an API could look like that accepts multiple encodings. Independently, just showing hex encoding in the UI (e.g. have a toggle) would be fairly easy.

ignatz:
Just FYI, added a select switch (on dev branch) to switch the blob encoding. You should now be able to grab a hex-encoded ID and use it in a SELECT.

ibilux:
> @ibilux thinking about this a little longer, this may be a caching issue, i.e. you might run the old UI against the new APIs. Could you try to "force" reload with Ctrl+Shift+R? (if that fixes it, I need to fix the ttls)

You are correct it's working fine now. The force reload with Ctrl+Shift+R fixed it (I have just learned something new tbh !).

Thank you @ignatz, and sorry for spamming this thread.

PR #187, opened 2025-12-09
docs: added uuid setting example

MentalGear:
standard sqlite does not have uuid, since trailbase adds it's own uuid function, it should be properly documented how to define it.


ignatz:
Much appreciated :pray: . I rebased your change in https://github.com/trailbaseio/trailbase/commit/ca50b8a28bc63758dcbb949601ca36e03bf5b470 (I also added a bit of verbiage) - Thanks!

PR #194, opened 2025-12-18
Added a register user api call to the rust client

pvg13:
I saw that the register user api call from `auth` did not have a function in the Rust client. I added it, although I had some issues when doing so. 

It seems like the API call for /api/auth/v1/register does not actually allow input in a JSON format, it requires the `CONTENT_TYPE: application/x-www-form-urlencoded` and the fields to be a form. 

I circumbeted the issues with a custom function that does the call with a new reqwest client however, probably there is a better solution in the server side that allows a json body and thus the register call can be unify in the same way as other API calls. 

I hope this helps!


ignatz:
Thanks and sorry it took so long. Register methods were added to all clients in: https://github.com/trailbaseio/trailbase/commit/3d457da4438ba70a3d38ee297b842d57a14a7600 :pray:

PR #204, opened 2025-12-30
Add qs serialization functionality, supporting encoding of filters and query parameters

a876691666:
This change is a missed addition. Using serialization and deserialization when intercepting query strings can be very helpful.

If possible, please push the new version of `trailbase-qs` to `crates.io` after merging

### DIFF
--- a/crates/qs/src/filter.rs
+++ b/crates/qs/src/filter.rs
@@ -19,53 +19,6 @@ pub enum Combiner {
   Or,
 }
 
-// --- Serialization helpers -------------------------------------------------
-
-pub(crate) fn encode_key(input: &str) -> String {
-  let mut out = String::with_capacity(input.len());
-  for b in input.as_bytes() {
-    let c = *b as char;
-    // unreserved: ALPHA / DIGIT / - . _ ~
-    if (c >= 'a' && c <= 'z')
-      || (c >= 'A' && c <= 'Z')
-      || (c >= '0' && c <= '9')
-      || c == '-'
-      || c == '.'
-      || c == '_'
-      || c == '~'
-    {
-      out.push(c);
-    } else {
-      // percent-encode everything else (including '[' and ']')
-      out.push_str(&format!("%{:02X}", b));
-    }
-  }
-  return out;
-}
-
-pub(crate) fn encode_val(input: &str) -> String {
-  let mut out = String::with_capacity(input.len());
-  for b in input.as_bytes() {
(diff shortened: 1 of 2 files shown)


a876691666:
Yes, I hope to implement a set of RBAC data permission filtering and retain the underlying filter support of Trailbase

Or to put it another way, my RBAC is built upon the Trailbase filtering layer, but currently, I can only implement it through the filter method. If there is a better API format, I will abandon this option

This is the advantage of Trailbase; otherwise, directly connecting to SQLite without any additional layers would be very troublesome

However, I would not choose Axum middleware

crates/qs/src/filter.rs
```diff
@@ -19,6 +19,53 @@ pub enum Combiner {
   Or,
 }
 
+// --- Serialization helpers -------------------------------------------------
+
+pub(crate) fn encode_key(input: &str) -> String {
```
ignatz: We should probably leave url encoding to an external library.

crates/wasm-runtime-guest/src/fetch.rs
```diff
@@ -1,20 +1,73 @@
...
+/// Note: to get raw bytes (Vec<u8>) use `fetch_bytes` or the convenience `get`.
+pub async fn fetch<T: DeserializeOwned>(
+  uri: impl ToString,
+  opts: FetchOptions,
+) -> Result<T, HttpError> {
+  let bytes = fetch_bytes(uri, opts).await?;
+
+  let parsed: T = serde_json::from_slice(&bytes)
+    .map_err(|err| HttpError::message(StatusCode::BAD_REQUEST, err))?;
+
```
ignatz: The response may not be json encoded.

ignatz:
To be mindful of your time and minimize churn, I added the Query serialization in (https://github.com/trailbaseio/trailbase/commit/974e8c54c10a517415ae2223c6b65bd8e73388a2 - basically your change rebased minus the wasm guest changes - Thanks :pray: ) and (https://github.com/trailbaseio/trailbase/commit/af87a2f516e5371ac0c622f89deb33b080d91961 - cleaning up a little, reducing API surface and maybe most importantly for your use-case removing the rusqlite dependency.).

We could add the the `post` function separately (arguably consistent with `get`), however none of this should be blocking.

Would you consider the current state workable? I can release a trailbase-qs crate.

a876691666:
I think this is great
Let's release a version😊

PR #205, opened 2025-12-30
Improve SPA fallback with Vite-style HTML resolution

takumi3488:
## Summary

- Add `--spa` CLI option to enable SPA fallback mode for `--public-dir`
- Vite-style HTML resolution: `/about` → `about.html` if exists
- Route requests (paths without file extensions) serve `index.html`
- File requests (paths with extensions) serve the file or return 404

## Motivation

When using client-side routing (React Router, Vue Router, TanStack Router, etc.),
refreshing the page on routes like `/user/profile` results in a 404 error because
no corresponding file exists on the server. This option enables the standard SPA
fallback pattern to handle these cases.

## Design Reference

This implementation is inspired by [Vite's HTML fallback middleware](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/server/middlewares/htmlFallback.ts),
with a production-oriented modification:

| Path | File Exists | Result |
|------|-------------|--------|
| `/about` | `about.html` exists | Serve `about.html` |
| `/about` | Not exists | Serve `index.html` (SPA) |
| `/docs/` | `docs/index.html` exists | Serve `docs/index.html` |
| `/style.css` | exists | Serve `style.css` |
| `/missing.css` | Not exists | **404** (differs from Vite) |

Unlike Vite's dev server (which always falls back to index.html), this implementation
returns 404 for missing static assets, which is more appropriate for production use.

### DIFF
--- a/crates/core/src/server/mod.rs
+++ b/crates/core/src/server/mod.rs
@@ -467,20 +467,36 @@ impl Server {
               }
               let contents = tokio::fs::read(file_path).await.ok()?;
               let mime = match file_path.extension().and_then(|e| e.to_str()) {
+                // Text
                 Some("html") => "text/html; charset=utf-8",
                 Some("css") => "text/css; charset=utf-8",
-                Some("js") => "application/javascript; charset=utf-8",
-                Some("json") => "application/json",
+                Some("js") | Some("mjs") => "application/javascript; charset=utf-8",
+                Some("json") | Some("map") => "application/json",
+                Some("txt") => "text/plain; charset=utf-8",
+                Some("xml") => "application/xml",
+                // Images
                 Some("png") => "image/png",
                 Some("jpg") | Some("jpeg") => "image/jpeg",
                 Some("gif") => "image/gif",
                 Some("svg") => "image/svg+xml",
                 Some("ico") => "image/x-icon",
+                Some("webp") => "image/webp",
+                Some("avif") => "image/avif",
+                // Video
+                Some("mp4") => "video/mp4",
+                Some("webm") => "video/webm",
+                // Audio
+                Some("mp3") => "audio/mpeg",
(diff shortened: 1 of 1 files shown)


takumi3488:
@a876691666 Thanks for sharing!

Personally, I think serving SPAs is a very common use case, so it should be supported with a simple option out-of-the-box. That said, if the project's direction is to handle this via custom code, I'm fine with that too!

takumi3488:
@ignatz Thanks for putting this together! Yes, your proposal works perfectly for my use case. As I mentioned earlier, my primary goal was the SPA fallback to index.html, and this achieves exactly that. The clean URL feature (/about → about.html) wasn't a hard requirement for me, so this simplified approach is more than enough.

Regarding the file extension handling (404 for missing assets): I still think it could be useful to have. Returning index.html for missing assets could make it harder to notice issues like typos in file paths or build failures. But again, this isn't a hard requirement — I'll leave it to your judgment.

Thanks for working on this!

ignatz:
> @ignatz Thanks for putting this together! Yes, your proposal works perfectly for my use case. As I mentioned earlier, my primary goal was the SPA fallback to index.html, and this achieves exactly that. The clean URL feature (/about → about.html) wasn't a hard requirement for me, so this simplified approach is more than enough.
> 
> Regarding the file extension handling (404 for missing assets): I still think it could be useful to have. Returning index.html for missing assets could make it harder to notice issues like typos in file paths or build failures. But again, this isn't a hard requirement — I'll leave it to your judgment.

My inner idealist would argue that file extensions are merely a convention but I think I'll have to agree with you. Serving index.html on  `/favicon.svg` would be a mess.

> Thanks for working on this!

Thank you for bringing this up. It was an over sight on my end that this isn't already implemented :pray:

Issue #208, opened 2026-01-11
Feature Request: OTP Authentication


wasantosfi:
Would be very interesting to have OTP Auth option considering:
- via **E-Mail** (using SMTP server, and easy integration for eg. with Google, Microsoft365, SendGrid, Hostinger, ProtonMail, Fastmail, etc);
- via **SMS** (using integrations for example with Twilio, TextMagic, Textla, etc.);
With possibility to set custom the expiration time (eg.: 5min, 15min, 20min, 30min, 1h, etc).

Possibility to serve as MFA resource for E-Mail + Password auth as well.

**Note**: As a first phase just alphanumeric OTP code is really an attractive resource (_mainly when we think on production-ready systems_), but later (very low priority for a while, maybe) magic link could be another resource to think as well.

Additionally, thank you very much for the great work!

ignatz:
Agreed. Second-factor auth has been on my short-list for a while. The only way to get two-factor at the moment is by deferring to a "social" OAuth or external OIDC auth provider.

ignatz:
@maxirozay that's awesome and divine timing, I was contemplating to start working on it.

> I would be happy to make a PR after getting some guidelines to make sure I do things well.

If you share the PR, we can discuss the details. I have a lot of questions, e.g. are you relying on email or are you supporting time-based OTPs, ... ?

🙏

ignatz:
Sounds all very reasonable 🙏 . If we wanted to support T-OTPs eventually (which we presumably want at some point), we certainly need some extra state in the DB. Safe travels

zyrakq:
https://github.com/trailbaseio/trailbase/pull/243 - fixed a bug when adding TOTP code

ignatz:
Thanks @zyrakq 🙏 . Forgot this one is still open. Should probably close even though the original FR was asking for SMS support too. If that's something we want we should probably track it separately. Not sure how standardized the integration (like SMTP) would be or whether we'd need an integration per provider.

PR #212, opened 2026-02-16
feat(OTP): add OTP

maxirozay:
This PR add OTP #208 by sending an email with a 8 characters code. It is implemented in the typescript client and in the admin login.

### DIFF
--- a/crates/assets/js/admin/proto/config.ts
+++ b/crates/assets/js/admin/proto/config.ts
@@ -346,6 +346,7 @@ export interface EmailConfig {
   userVerificationTemplate?: EmailTemplate | undefined;
   passwordResetTemplate?: EmailTemplate | undefined;
   changeEmailTemplate?: EmailTemplate | undefined;
+  otpTemplate?: EmailTemplate | undefined;
 }
 
 export interface OAuthProviderConfig {
@@ -702,6 +703,9 @@ export const EmailConfig: MessageFns<EmailConfig> = {
     if (message.changeEmailTemplate !== undefined) {
       EmailTemplate.encode(message.changeEmailTemplate, writer.uint32(186).fork()).join();
     }
+    if (message.otpTemplate !== undefined) {
+      EmailTemplate.encode(message.otpTemplate, writer.uint32(194).fork()).join();
+    }
     return writer;
   },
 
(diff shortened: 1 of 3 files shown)


ignatz:
Sorry, for the radio silence - apologies! I fell down a cross-platform-build rabbit hole and must have banged my head.

I'm back now and will look at this ASAP - thanks already :pray: 

> In my company we switched from password auth to passwordless to have less support and firebase magic link works very well. I think it's good to have the option and for sensitive accounts use multifactor auth. Password manager are good but not everybody uses them and clients are happy to just receive a link or code rather than have to remember a password that they will forget and then do the "forget password" flow to sign in every time ^^.

Thanks for the motivation. I guess I could see this work well for internal tooling :+1: . I'll also take a closer look at what firebase offers (and supabase apparently).

> The OTP with 12 characters that last 5 minutes can be a sign in option alone. It's basically a 12 characters random password that will be reset in 5 minutes....Do you have preferences or suggestions?

W/o yet having looked at it, I would assume we could do many digits. Isn't it just a link folks click on?

> We return an auth code like in login_with_authorization_code_flow_and_pkce and ask for the auth code plus the TOTP to sign in.

That sounds pretty much what I would have naively expected :+1: 

> Sumary of what I did:
>
> *    Sign in with OTP
> *    Sign in with password plus TOTP
> *    Sign in with OTP plus TOTP

That's amazing :tada: :pray:

ignatz:
> I don't know what can be done with the auth UI but if that can help, on the project i previously linked, I made a route to get all the sign in options. I only ask for the email then I get the options and show them to the user. So if a user has TOTP after validating their email I'll show the TOTP field.

Appreciated. This works well for client-side JS. Ideally all auth UIs would work w/o JS purely relying on HTML forms and SSR. At the same time the auth APIs and UIs need to be decoupled.

I ended up passing an mfa_redirect target to the login, so the API can hand back to the UI, when TOTP or other MFA is required, which can then pass an mfa_token plus second secret to a mfa_login endpoint. Not sure this is most elegant but it works :/.

This is all turning out to be pretty big :hide:

ignatz:
Quick update. I have most parts in place including TOTP and OTP and a bunch of nice improvements. Besides some more testing and updating more clients, I'm primarily missing a good way to pass configuration/state to the admin SPA, specifically if OTP is enabled or not for the instance. I would either:

* Have to change how we render/deploy the admin UI
* Risk showing a disfunctional button (which may be ok)
* Not allow OTP login through the admin login page (FWIW, it would still work to log in through the auth component and then go the admin UI :/)

ignatz:
v0.25.0 adds second-factor TOTP and OTP email support in the server, the two UIs and all 8 clients. It also changes how we handle session like state and OTP codes are moved to a separate `session.db`. Thanks @maxirozay :pray:

PR #232, opened 2026-05-12
fix: skip stored FileUpload format in PATCH requests

### LINKED ISSUE #233: PATCH /api/records/v1/<table>/<id> fails with 400 when request includes a stored FileUpload reference
## Description

When a client reads a record that has a `std.FileUpload` or `std.FileUploads` column and then sends a PATCH request that includes the **stored** file reference format (the JSON returned by GET — which has `filename`, `mime_type`, etc. but **no** `data` field), the request fails with `400 invalid params`.

## Reproduction

```
# 1. Upload a file to a record
POST /api/records/v1/user_profiles
{ ..., "avatar": { "filename": "photo.jpg", "content_type": "image/jpeg", "data": "<base64>" } }
→ 200

# 2. Read the record back — stored format has no 'data' field
GET /api/records/v1/user_profiles/{id}
→ { "avatar": { "filename": "photo_xyz.jpg", "mime_type": "image/jpeg", ... } }

# 3. Send PATCH with other fields + echoed avatar reference
PATCH /api/records/v1/user_profiles/{id}
{ "display_name": "Alice", "avatar": { "filename": "photo_xyz.jpg", "mime_type": "image/jpeg", ... } }
→ 400 "invalid params"
```

## Root cause

`extract_params_and_files_from_json` in `crates/core/src/records/params.rs` attempts to deserialize any `std.FileUpload` value as `FileUploadInput`, which has:
- `#[serde(deny_unknown_fields)]` — rejects `mime_type`, `original_filename`, `id` from the stored format
- `data: Vec<u8>` — required, non-optional

When a client (e.g. TanStack DB, any framework that round-trips records) sends back the stored file reference, deserialization fails with a JSON error → `400 invalid params`.

## Expected behavior

A PATCH request that includes the stored `FileUpload` format should be treated as **"keep existing file"** — the same semantics as omitting the field entirely. Only when a `data` field is present should the file be (re-)uploaded.

## Impact

This affects any client that:
- Reads a record containing a file upload
- Sends a PATCH with only some fields changed (e.g. updating a display name without changing the avatar)
- Includes the full current record values in the PATCH body — a common pattern in optimistic-update frameworks

## Fix

PR #232 adds a pre-check in `Params::for_update` to detect the stored format (absence of `data` key) and skip the column, preserving the existing file.

royalcala:
Fixes #233

## Problem

When a client performs a PATCH request where a `std.FileUpload` or `std.FileUploads` column contains the **stored** file reference format (the JSON returned by a GET — which has `filename`, `mime_type`, etc. but **no** `data` field), the request fails with `400 invalid params`.

This happens because `extract_params_and_files_from_json` tries to deserialize the value as `FileUploadInput`, which has `#[serde(deny_unknown_fields)]` and requires a non-optional `data: Vec<u8>` field. The stored `FileUpload` format has `mime_type` (unknown field) and no `data`, so deserialization fails.

### Reproduction

```
GET /api/records/v1/user_profiles/{id}
→ { "avatar": { "filename": "abc.jpg", "mime_type": "image/jpeg", ... } }

PATCH /api/records/v1/user_profiles/{id}
  { "display_name": "New Name", "avatar": { "filename": "abc.jpg", "mime_type": "image/jpeg", ... } }
→ 400 "invalid params"
```

This is a common pattern for any client that reads a record and sends back a partial update (e.g. TanStack DB, which sends only changed fields but includes the full current values of unchanged fields in the `changes` delta).

## Fix

In `Params::for_update`, before calling `extract_params_and_files_from_json`, detect the stored format by checking for the **absence of a `data` field** in the JSON value:

- `std.FileUpload`: if value is an Object without a `"data"` key → skip the column (keep existing file)
- `std.FileUploads`: if value is an Array where all items are Objects without `"data"` keys → skip the column (keep existing files)

When skipped, the column is simply not included in the UPDATE statement, so the existing file reference in the database is preserved unchanged.

This makes PATCH semantics consistent: fields absent from the PATCH body are left unchanged, and file fields echoed back in their stored format behave the same way.

## Invariants preserved

- **INSERT** (`for_insert`): unchanged — new uploads still require `data`
- **PATCH with new file** (`for_update` + `data` present): unchanged — new file upload replaces existing
- **PATCH without file field**: unchanged — column not in body is not updated
- **PATCH with stored format** (new behavior): column is skipped, existing file preserved

## Test

Added `test_for_update_skips_stored_file_upload` in `crates/core/src/records/params.rs` that verifies:
1. PATCH with stored `FileUpload` format (no `data`) → column skipped, no error
2. PATCH with new `FileUploadInput` (with `data`) → column included, file uploaded

### DIFF
--- a/crates/core/src/records/params.rs
+++ b/crates/core/src/records/params.rs
@@ -266,6 +266,23 @@ impl Params {
         continue;
       };
 
+      // For file upload columns in PATCH requests: if the client echoes back the stored
+      // FileUpload/FileUploads format (no 'data' field), skip the column entirely rather
+      // than returning an error. This preserves the existing file without modification,
+      // which is the expected behavior when a client round-trips a record through the API.
+      if let Some(JsonColumnMetadata::SchemaName(schema_name)) = json.as_ref() {
+        let is_stored_upload = match (schema_name.as_str(), &value) {
+          ("std.FileUpload", serde_json::Value::Object(map)) => !map.contains_key("data"),
+          ("std.FileUploads", serde_json::Value::Array(items)) => items
+            .iter()
+            .all(|v| matches!(v, serde_json::Value::Object(m) if !m.contains_key("data"))),
+          _ => false,
+        };
+        if is_stored_upload {
+          continue;
+        }
+      }
+
       let (param, json_files) = extract_params_and_files_from_json(
         json_schema_registry,
         column,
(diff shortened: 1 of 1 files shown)


ignatz:
This change was rebased as https://github.com/trailbaseio/trailbase/commit/bb76986bfccc95e863a74e838b7132dca131b40f. Much appreciated :pray:

PR #235, opened 2026-05-19
Fix spelling in `pocketbase.mdx`

naltun:
I was reviewing the docs and came across this misspelling, so I'm sending this patch over in case you find it useful.

Great project!


ignatz:
Appreciated :pray:

PR #241, opened 2026-06-03
fix(config): validate smtp_host is a hostname or IP

Kartalops:
## What

validate_email_config accepted any non-empty smtp_host string. A typo like `smtp_host=mail.example.com:587` (with port) or `smtp_host=http://mail.example.com/` would propagate through to the SMTP client and fail at connect time with a less actionable error. The TODO at the top of the function asked for this check explicitly.

## Why

A misconfigured smtp_host should fail at config load with a clear ConfigError pointing at the offending value, which is the same code path other config issues already use. The existing TODO marker pinned this as a known gap.

## How

- Rename the captured host from `_host` to `host` and reject empty values up front.
- Add an `is_valid_smtp_host` helper that accepts strings that parse as `std::net::IpAddr` or match a strict hostname character set (alnum, '.', '-') with at least one dot.
- Reject the new helper's output with a ConfigError.

The strict set rejects common typos (port, scheme, path) and is consistent with SMTP host semantics: bare hostname (or IP) only.

## Notes

- Could not run `cargo check` here (no Rust toolchain in this environment); the change is isolated to one file and only adds new error paths.
- Single-label hostnames (e.g. `intranet`) are intentionally rejected. Happy to relax to allow these if maintainers prefer.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -683,7 +683,7 @@ pub(crate) fn validate_email_config(email: &proto::EmailConfig) -> Result<(), Co
   validate_email_template(email.password_reset_template.as_ref())?;
   validate_email_template(email.otp_template.as_ref())?;
 
-  let Some(_host) = &email.smtp_host else {
+  let Some(host) = &email.smtp_host else {
     match (email.smtp_port, &email.smtp_username, &email.smtp_password) {
       (None, None, None) => {
         // No SMTP configured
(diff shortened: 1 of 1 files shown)


ignatz:
Hey @Arvuno , thanks for the PR and stepping on my toes on the missing validation directly addressing the pending TODO.

I addressed the issue in https://github.com/trailbaseio/trailbase/commit/5963080628c8b47a0d4e6e1f4c6dd8d34e15abee#diff-57cbbec45e799304c658c9dea8b670297b898fab012b543fcbb266e0b06fa42fR812, given the LLM approach seemed a little bit verbose on what should simply be:

```rust
url::Host::parse(host).is_ok()
```

Still very much appreciate the request :pray:

PR #242, opened 2026-06-10
[draft] fix: allow use as library dependency without git submodules

zyrakq:
## Problem

When TrailBase is consumed as a library from an external workspace (e.g. via `git subtree`), the build fails because:

1. `sqlite-vec` and `pgrow2serde` are path dependencies pointing to unpopulated git submodules.
2. `serde_rusqlite` uses `rusqlite = { workspace = true }`, which requires TrailBase to be the active workspace root — this breaks when resolved from outside.

## Solution

Replace the three sub-dependencies with sources that work standalone, while keeping `[patch]` entries so local submodule paths are still used when building TrailBase directly.

Since `[patch]` is only applied from the root workspace, contributors with submodules populated see no change in behavior.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,8 +71,8 @@ rust-embed = { version = "8.4.0", default-features = false, features = ["mime-gu
 serde = { version = "^1.0.203", features = ["derive", "rc"] }
 serde_json = { version = "^1.0.117" }
 serde_qs = { version = "1.0.0", default-features = false, features = [] }
-serde_rusqlite = { path = "vendor/serde_rusqlite" }
-sqlite-vec = { path = "vendor/sqlite-vec/bindings/rust", default-features = false }
+serde_rusqlite = "0.43.0"
+sqlite-vec = { git = "https://github.com/ignatz/sqlite-vec", rev = "main", default-features = false }
 sqlite3-parser = "0.16.0"
 tokio = { version = "^1.38.0", default-features = false, features = ["fs", "io-std", "macros", "net", "rt-multi-thread", "signal", "sync", "time"] }
 tokio-rustls = { version = "0.26.1", default-features = false }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -71,8 +71,8 @@ rust-embed = { version = "8.4.0", default-features = false, features = ["mime-gu
 serde = { version = "^1.0.203", features = ["derive", "rc"] }
 serde_json = { version = "^1.0.117" }
 serde_qs = { version = "1.0.0", default-features = false, features = [] }
-serde_rusqlite = { path = "vendor/serde_rusqlite" }
-sqlite-vec = { path = "vendor/sqlite-vec/bindings/rust", default-features = false }
+serde_rusqlite = "0.43.0"
```
ignatz, line 74: (No action required)

serde_rusqlite is a well-maintained crate. I merely vendored it in the past since sometimes it would take a few weeks before rusqlite updates propagate. On the flip-side my vendored version has become stale, so thank you for pointing it back at upstream.

ignatz:
This looks great, thanks! Will squash as soon as the workflows pass.

Out of curiosity, I suspect you're using TB as a library - is that right? For context, with APIs internally changing quite a bit (e.g. lately DB init sync -> async to accommodate for PG) this is something I haven't payed too much attention to. Appreciate you not being discouraged by these obstacles. Please keep it coming :pray:

zyrakq:
I was a bit hasty in opening the PR and specified some inappropriate versions of the packages. The test will likely fail, but even if it does, something might break. Unfortunately, I'm having some issues with the network and development environment, so I won't be able to fix it until tomorrow.

ignatz:
> I was a bit hasty in opening the PR and specified some inappropriate versions of the packages. The test will likely fail, but even if it does, something might break. Unfortunately, I'm having some issues with the network and development environment, so I won't be able to fix it until tomorrow.

I suspect it's only an issue for you (i.e. I expect tests and workspace builds to pass), specifically `vendor/sqlite-vec` should be pointed at the `v0.1.10-alpha.4-fork` branch.  Happy to fix up.

ignatz:
Just pushed it to the dev branch: https://github.com/trailbaseio/trailbase/commit/e91c17e3280dcf5786b01c57ce8298091eb7d7b5 :pray: 

Will rebase a few more minor changes and release a new version to get the OTP fix out ASAP. Thanks again

zyrakq:
Yes, I'm using TB as a library 👍
I haven't run into any issues so far, except for my own mistakes with package versions.  

I actually got started with Trailbase as a BaaS, so I'm not very up to speed on the internal challenges you're dealing with (like the recent DB init sync → async changes). Thanks for the context.

I skimmed through some of the open issues — a few of the features discussed there are things I’d also like to see. Might try to propose something in the future around making the admin panel more extensible (for example, via web components and micro-frontends).

PR #243, opened 2026-06-11
fix: remove escaped quotes in register_totp_confirm UPDATE query

zyrakq:
auth/api/totp.rs` used `\"` inside a raw string literal for the `UPDATE _user SET totp_secret` query. SQLite received literal backslash-quote characters instead of column quotes, causing the query to fail with a syntax error when enabling TOTP.

Change the `UPDATE_QUERY` constant in `register_totp_confirm_handler` to use unquoted
double-quotes (`"`) consistent with the adjacent `unregister_totp_handler`.

https://github.com/trailbaseio/trailbase/issues/208

### DIFF
--- a/crates/core/src/auth/api/totp.rs
+++ b/crates/core/src/auth/api/totp.rs
@@ -111,7 +111,7 @@ pub async fn register_totp_confirm_handler(
   }
 
   const UPDATE_QUERY: &str =
-    formatcp!(r#"UPDATE \"{USER_TABLE}\" SET totp_secret = $1 WHERE id = $2"#);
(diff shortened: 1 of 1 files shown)


ignatz:
Yikes. Thank you :pray:

For context, this  must have happened during the PG transition. Previoulsy, we we're using `'` in many places since they don't require escaping in rust strings, neither in standard nor raw strings but `"` somestimes do. Anyway, I quickly audited all the other occurrences of `\"` in raw strings and this is the only instance I found :pray:

I'll also look into improving test coverage.

PR #249, opened 2026-06-17
Fix password-reset email template validation

zyrakq:
## Problem

`validate_email_template` in `config.rs` applied a single generic check to all four template types, requiring every non-empty body to contain `{{ VERIFICATION_URL }}` or `{{ CODE }}`.

The password-reset renderer (`password_reset_email`) never provides `VERIFICATION_URL` — it builds the link from `{{ SITE_URL }}` + `{{ TOKEN }}`. As a result:

- Saving any password-reset template body from the admin UI failed validation with a misleading `VERIFICATION_URL` error.
- Following that error and adding `{{ VERIFICATION_URL }}` made validation pass but **broke rendering at send time** (undefined variable → template error).

## Fix

Parameterize `validate_email_template` with per-type acceptable variables, so each template validates against what its renderer actually provides:

| Template | Acceptable vars |
|---|---|
| Email Verification | `VERIFICATION_URL`, `CODE`, `TOKEN` |
| Change Email | `VERIFICATION_URL`, `CODE`, `TOKEN` |
| Password Reset | `TOKEN`, `CODE` |
| OTP | `CODE` |

The error message now lists the acceptable variables for the specific template. Empty/None bodies still fall back to defaults. Unit tests cover each case.

### DIFF
--- a/crates/core/src/config.rs
+++ b/crates/core/src/config.rs
@@ -678,10 +678,19 @@ pub async fn validate_config(
 }
 
 pub(crate) fn validate_email_config(email: &proto::EmailConfig) -> Result<(), ConfigError> {
-  validate_email_template(email.user_verification_template.as_ref())?;
-  validate_email_template(email.change_email_template.as_ref())?;
-  validate_email_template(email.password_reset_template.as_ref())?;
-  validate_email_template(email.otp_template.as_ref())?;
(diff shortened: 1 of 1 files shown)


ignatz:
Much appreciated :pray: . Pulled your change verbatim into the dev branch

PR #250, opened 2026-06-17
Fix /status endpoint not updating auth cookie after refresh

zyrakq:
## Problem

`GET /status` rotates the JWT (new `csrf_token`) but does not update the `auth_token` cookie. The response body and the cookie permanently diverge. Cookie-based clients get `400 Invalid CSRF token` on the next CSRF-protected request (e.g. `/change_email/request`), because the body `csrf_token` ≠ cookie `csrf_token`.

## Root cause

`login_status_handler` calls `reauth_with_refresh_token()` but never calls `cookies.add()`. The `Tokens` extractor (`tokens.rs:99-108`) does the same reauth and correctly sets the cookie. The status handler simply forgot to do the same.

## Fix

Add `cookies: Cookies` to `login_status_handler` and call `cookies.add(new_cookie(...))` after reauth — mirroring what `extract_tokens_from_request_parts` already does for stale auth cookies.

### DIFF
--- a/crates/core/src/auth/api/status.rs
+++ b/crates/core/src/auth/api/status.rs
@@ -1,11 +1,14 @@
 use axum::extract::{Json, State};
 use serde::{Deserialize, Serialize};
+use tower_cookies::Cookies;
 use ts_rs::TS;
 use utoipa::ToSchema;
 
 use crate::app_state::AppState;
 use crate::auth::AuthError;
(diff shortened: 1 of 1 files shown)


ignatz:
Much appreciated :pray: 

I pushed your change to the dev-branch. I made the cookie update conditional and made the formatter happy.

PR #251, opened 2026-06-17
feat(auth-ui): add profile API endpoint for per-user capabilities

zyrakq:
**Description:**

Adds a new `GET /api/auth/v1/profile` endpoint to the auth-ui WASM component that returns server-side capability flags for the currently authenticated user.

**Why:** The profile page unconditionally shows the TOTP setup/disable button and the "Change Password" link — including for users who signed in via an external OAuth/OIDC provider. OAuth users have no password, so both options are misleading and non-functional for them.

**What changed:**
- New `GET /api/auth/v1/profile` route in `auth-ui` WASM. Requires authentication (401 if not logged in). Queries `provider_id` and `password_hash` from the `_user` table and reads `config:auth` from the KV store to compute two capability flags. A user is considered OAuth-only when `provider_id != 0` and `password_hash` is empty; users with an admin-assigned password are treated as password-capable regardless of `provider_id`.
- `Profile.tsx` uses `createResource` to fetch this endpoint, wraps `<TotpToggleButton>` in `<Show when={showOtpSection()}>`, and wraps the "Change Password" link in `<Show when={showChangePassword()}>`.

**Response shape:**
```json
{ "show_otp_section": true, "show_change_password": false }
```

Both flags encode the full server-side decision, so clients don't need to reason about `provider_id`, `password_hash`, or config fields separately.

### DIFF
--- a/crates/auth-ui/src/lib.rs
+++ b/crates/auth-ui/src/lib.rs
@@ -3,9 +3,10 @@
 #![warn(clippy::await_holding_lock, clippy::inefficient_to_string)]
 
 use askama::Template;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
+use trailbase_wasm::db;
 use trailbase_wasm::http::{
-  Html, HttpError, HttpRoute, IntoBody, IntoResponse, Redirect, Request, Response, StatusCode,
+  Html, HttpError, HttpRoute, IntoBody, IntoResponse, Json, Redirect, Request, Response, StatusCode,
   User, header, routing,
 };
 use trailbase_wasm::kv::Store;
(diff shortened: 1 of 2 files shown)


ignatz:
Your report makes a lot of sense :pray: . I'm inclined to add the auth type to the JWT token and pass the information that way. WDYT? Would that work for you?

ignatz:
Should be addressed in https://github.com/trailbaseio/trailbase/commit/5e0c5c989b057f5122b6891d5719ef13c5a1124b. I'll close the PR for now but happy to re-open. Thanks again :pray:

zyrakq:
Sounds good, thanks for looking into it 🙏

Not a big deal for me either — I'm using my own auth module anyway, so the JWT change works fine. Your approach is cleaner overall. I went with the separate endpoint mostly to avoid touching Trailbase core, plus I first wrote this bit for my own module where the empty-hash check is still useful (users can set a password later from the profile, which the JWT-only path wouldn't support). 

Happy either way!

PR #254, opened 2026-06-24
Fix spurious scrollbar on accounts page

zyrakq:
Removed h-full from the AccountsPage root div to prevent sub-pixel height overflow from producing a spurious scrollbar Added overflow-y-auto to the TopNav main element so tall content scrolls inside the panel instead of the document body

AccountsPage produced a scrollbar even when content was shorter than the viewport due to h-dvh vs 100% sub-pixel rounding in the browser


ignatz:
Much appreciated :pray: . Cherry-picked this change verbatim into the dev branch: https://github.com/trailbaseio/trailbase/commit/7b26aaf21bc9aa7f6f55a94dbade4a858ae29bea

PR #256, opened 2026-06-29
Update custom-binary example to trigger migrations

bltavares:
The `example/custom-binary` uses the `trailbase::api::serve` function,
which starts the services but don't trigger the boostrap user migrations
itself.

Reverse enginering how the `cli::run` command works, it seems it's
triggered by `Server::serve` function instead.

This commit updates the example code to use the `Server:serve` function
to ensure initial migration logic is included on the custom binary.

---

Note: the current example fails in `debug` mode, as it attempts to access
`User` from the `trailbase` extractor, but there is an `assert!` informing that
the `CookieManage` layer is not installed.

It should be possible to fix it by adding the `CookieManage` layer as well on
the custom router, but maybe a better fix would be to pass in an
`custom_router` on `ServerOptions` to be merged and use the same shared layers.

### DIFF
--- a/examples/custom-binary/src/main.rs
+++ b/examples/custom-binary/src/main.rs
@@ -45,12 +45,7 @@ async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
     .install_default()
     .expect("Failed to install rustls crypto");
 
-  let Server {
-    state,
-    main_router,
-    admin_router,
-    tls,
-  } = Server::init_with_custom_initializer(
+  let server = Server::init_with_custom_initializer(
     ServerOptions {
       data_dir: DataDir::default(),
       address: "localhost:4004".to_string(),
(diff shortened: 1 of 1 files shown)


ignatz:
I'm sympathetic to bootstrapping user-defined migrations and reducing the divergence between Server::serve and api::serve. However, as it stands an example that doesn't work is worse. It's not trivial to just inject a router due to the dependence of AppState via CustomState. We'd have to inject a builder (with all the generic nonsense that it brings to ServerOptions) or we'd need to pull out the AppState consruction or we'd need to add another Server livecycle, or ....

Let me sleep over it.

bltavares:
Thanks for replying. No need to rush on it, and feel free to close it also. I was more keen to report the finding of migrations not being triggered by the example in case someone else attempts it.

ignatz:
> Thanks for replying. No need to rush on it, and feel free to close it also

I appreciate it when folks step on my toes :pray:. I went the long (and hopefully most principled) way of pulling out the AppState construction. Take a look at https://github.com/trailbaseio/trailbase/blob/dev/examples/custom-binary/src/main.rs#L48

bltavares:
Nice! I like the proposed API, seems much easier to add routers without exposing the `tls` and named router tuple to the API

PR #259, opened 2026-07-06
fix: bundle TypeScript types as expected

bltavares:
The project is using `vite-plugin-dts` to prepare the JS/TS library for
publishing functions with type annotations. The `package.json` expects a
`index.d.ts` file to exist to offer types to clients.

The upgrade to `vite-plugin-dts` to v5 had a breaking change on the API,
which now requires `@microsoft/api-extractor` to produce a unified
`index.d.ts` instead of a separate `src/*.d.ts`. To run this library,
`ajv` was also required as an optional peer package (at least on Bun).

The breaking change on the `vite-plugin-dts` also had some config field
name changes which has adjusted.

To avoid regressions, `npm run prepack` will check for the `index.d.ts`
file in case there are future breaking changes.

Without this file, downstream clients cannot find types:

```text
error: Lint or type issues found
× typescript(TS7016): Could not find a declaration file for module 'trailbase'. 'F:/repos/vintage/node_modules/trailbase/dist/index.js' implicitly has an 'any' type.
   ╭─[src/routes/product/[id]/+page.ts:2:29]
 2 │ import { type Client } from "trailbase";
 ```

### DIFF
--- a/crates/assets/js/client/package.json
+++ b/crates/assets/js/client/package.json
@@ -30,7 +30,7 @@
     "build": "vite build",
     "check": "tsc --noEmit --skipLibCheck && eslint",
     "format": "prettier -w src tests",
-    "prepack": "rm -rf ./dist && pnpm build && test -e ./dist/index.js",
+    "prepack": "rm -rf ./dist && pnpm build && test -e ./dist/index.js && test -e ./dist/index.d.ts",
     "test": "vitest run && vite-node tests/integration_test_runner.ts"
   },
   "dependencies": {
@@ -42,6 +42,8 @@
   },
   "devDependencies": {
     "@eslint/js": "^10.0.1",
+    "@microsoft/api-extractor": "^7.58.9",
+    "ajv": "^8.20.0",
     "eslint": "^10.4.1",
     "globals": "^17.6.0",
     "http-status": "^2.1.0",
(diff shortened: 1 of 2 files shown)


ignatz:
Much appreciated :pray: . I was trying to understand what is broken so I can merge and publish this ASAP. I had quick look at the latest package: 
Screenshot From 2026-07-07 17-53-11

and it seems to export the types such as aforementioned `Client`. Is this an issue with the published package are you doing something else to trigger the issue?

ignatz:
Figured it out: both the path is off and types don't get rolled up :pray: . Submitted your change in https://github.com/trailbaseio/trailbase/commit/686fccda6380980986dc8aa1112f519d555c3282 (minor changes: bump version, remove dep and fix the other place rollupTypes was referenced). Thanks so much :pray:

ignatz:
Just published a new npm v0.13.1 :pray:

bltavares:
Sorry for the delay responding to it, I spent a few days traveling. I'm glad you found it, but for the record:

- The `package.json` declares types as bundled `index.d.ts` instead of folder with `*.d.ts`
https://github.com/trailbaseio/trailbase/blob/686fccda6380980986dc8aa1112f519d555c3282/crates/assets/js/client/package.json#L16
- It used to be published as `index.d.ts` but the breaking change upgrade to the `vite-plugin-dts` left the defaults to output a folder with `*.dt.s` instead

The `ajv` dep might have been an artifact of using Bun and some issues with peer dependencies, definitely not needed to publish.

Thank you for fixing. I can confirm the latest release picks up the definition.

PR #262, opened 2026-07-15
Add native authenticated MCP support to TrailBase

brigon-dev:
## Summary

Adds optional, native MCP support directly to the TrailBase binary and existing Docker container. MCP is served from `/mcp` on the TrailBase admin server and is enabled explicitly with `--mcp`.

This replaces the earlier FastMCP sidecar design. There is no second container, second port, copied bearer token, shared depot mount, or hand-maintained public API catalog.

## Authentication and security

- Browser-based OAuth authorization code flow with PKCE S256
- OAuth protected-resource and authorization-server discovery metadata
- Dynamic client registration for clients such as `mcp-remote`
- Access-token refresh
- MCP tokens scoped to `mcp` and audience-bound to the instance `/mcp` URL
- Live administrator-status lookup on every MCP request
- Existing administrator bearer tokens supported as an explicit compatibility mode
- MCP disabled by default

## Tools

- `call_admin_api`: dispatches to TrailBase existing in-process admin router to avoid API drift
- `execute_sql`: runs SQL using TrailBase administrative query handling
- `list_tables`: reads current tables, views, indexes, and triggers
- `get_config`: returns redacted TrailBase configuration
- `update_config`: validates and updates configuration while preserving secrets

This gives trusted administrators the same administrative surface used by the dashboard, including schema, table, Record API, user, file, backup, job, OAuth-provider, and WASM-component operations.

## Deployment

```sh
trail --public-url https://trailbase.example.com run --mcp
```

OAuth-capable clients connect to `https://trailbase.example.com/mcp`. The documentation includes localhost, `mcp-remote`, Docker, Portainer, Cloudflare Tunnel, reverse-proxy, and direct bearer-token examples.

## Validation

- Native MCP unit tests pass
- TrailBase CLI build and checks pass
- Browser OAuth, PKCE callback, token exchange, and refresh flow tested against a disposable local depot
- MCP initialization and tool discovery tested with RustRover through `mcp-remote`
- Table creation, row insertion, schema listing, and SQL reads tested through MCP
- No personal or production TrailBase data was modified during destructive testing

This redesign addresses the review feedback by moving MCP into the main binary, routing operations through existing admin handlers, avoiding a separately maintained operation catalog, and supporting administrative schema changes.


.cargo/config.toml
```diff
@@ -14,3 +14,8 @@
 
 [env]
 TS_RS_EXPORT_DIR = { value = "./crates/assets/js/bindings", relative = true }
+LIBCLANG_PATH = { value = "./.dev-tools/libclang-18/usr/lib/llvm-18/lib", relative = true }
+CLANG_PATH = { value = "./.dev-tools/libclang-18/usr/bin/clang-18", relative = true }
+PKG_CONFIG_PATH = { value = "./.dev-tools/geos/usr/lib/x86_64-linux-gnu/pkgconfig", relative = true }
+PKG_CONFIG_SYSROOT_DIR = { value = "./.dev-tools/geos", relative = true }
+PROTOC = { value = "./.cargo/protoc-wrapper.sh", relative = true }
```
ignatz, line 21: I'm going to assume that his is an artifact

.cargo/protoc-wrapper.sh
```diff
@@ -0,0 +1,11 @@
...
+set -eu
+
+SCRIPT_DIR=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+REPO_ROOT=$(CDPATH= cd -- "${SCRIPT_DIR}/.." && pwd)
+PROTOBUF_DIR="${REPO_ROOT}/.dev-tools/protobuf"
+
+export LD_LIBRARY_PATH="${PROTOBUF_DIR}/usr/lib/x86_64-linux-gnu${LD_LIBRARY_PATH:+:${LD_LIBRARY_PATH}}"
+exec "${PROTOBUF_DIR}/usr/bin/protoc" \
+  -I"${PROTOBUF_DIR}/usr/include" \
+  "$@"
```
ignatz, line 11: Same as above?

README.md
```diff
@@ -162,6 +177,15 @@ pnpm install
...
 ```
 
+On Debian/Ubuntu-style systems without sudo access to install build packages,
+this fork can populate the repo-local `.dev-tools/` cache used by
+`.cargo/config.toml`:
+
+```sh
+scripts/bootstrap-local-dev-tools.sh
+cargo check --workspace --all-targets
+```
```
ignatz, line 186: Ok, so the above files are not artifacts.

This is a bit surprising since, I would expect most devs to have access to their machines (seems like a reasonable requirement). Otherwise, this is also very deb centric. If you don't have full access, wouldn't one rather develop inside a container? Would love to hear more about the reasoning.

mcp/src/trailbase_mcp/endpoints.py
```diff
@@ -0,0 +1,370 @@
...
+TRAILBASE_API_OPERATIONS: tuple[dict[str, Any], ...] = (
+    {
+        "operation_id": "auth_code_to_token_handler",
+        "category": "auth",
+        "method": "POST",
+        "path": "/api/auth/v1/token",
+        "summary": "Exchange authorization code for auth tokens.",
+        "mcp_support": "call_trailbase_api_operation or trailbase_request",
+        "requires_write_permission": True,
+    },
```
ignatz: Just guessing, wouldn't the MCP also need a POST request definition?

mcp/src/trailbase_mcp/endpoints.py
```diff
@@ -0,0 +1,370 @@
+from __future__ import annotations
+
+from typing import Any
+from urllib.parse import quote
+
+TRAILBASE_API_OPERATIONS: tuple[dict[str, Any], ...] = (
```
ignatz: I'm a bit worried that this will get out of sync. Naively, I would have expected the MCP implementation to be part of the main binary running in some dev mode.

mcp/src/trailbase_mcp/endpoints.py
```diff
@@ -0,0 +1,370 @@
...
+    {
```
ignatz: I'm also a bit surprised over the selection of methods. Naively, I would have expected only or mostly admin APIs to be used in dev mode basically as an alternative to the dashboard. Isn't exposing only the public APIs with access protection inherently limiting maybe even useless for dev tasks.

As an example, i would have expected this to be used to drive schema changes.

ignatz (commented):
Thanks jumping into the cold water - much appreciated 🙏 

To reduce churn, maybe its best if first work out some of the highlevel questions.

brigon-dev:
Thanks again for the earlier feedback @ignatz. I’ve substantially redesigned this PR around your suggestions.

The MCP server is now part of the main TrailBase binary and runs in the same container and process as TrailBase. It is enabled explicitly with `--mcp` and served from `/mcp` on the existing admin server, so there is no longer a FastMCP sidecar, second container, second port, shared depot mount, copied bearer token, or separately maintained API-operation catalog.

Authentication now uses a browser-based OAuth authorization-code flow with PKCE. MCP clients open TrailBase’s existing login UI, and only current administrators are authorized. MCP access tokens are scoped and audience-bound to the instance’s `/mcp` URL.

MCP operations are routed through TrailBase’s existing in-process admin handlers to reduce drift and provide the administrative functionality discussed in the review, including schema changes, tables, indexes, triggers, Record APIs, configuration, users, files, backups, jobs, OAuth providers, and WASM components. Focused tools are also included for SQL, schema discovery, and configuration.

I tested the complete flow with a disposable local TrailBase instance and RustRover via `mcp-remote`: browser login, PKCE callback, token exchange, MCP initialization, tool discovery, table creation, row insertion, and reads all succeeded.

The PR title and description have been updated to reflect the new implementation. I’d appreciate another review when you have time, particularly around whether the native integration and admin-tool approach now align with what you had in mind.

ignatz:
> The MCP server is now part of the main TrailBase binary and runs in the same container and process as TrailBase. It is enabled explicitly with `--mcp` and served from `/mcp` on the existing admin server, so there is no longer a FastMCP sidecar, second container, second port, shared depot mount, copied bearer token, or separately maintained API-operation catalog.

:pray: gave it a quick skim and it looks much closer to what I had in mind. I still have a few questions, some of it may just be my ignorance showing.

> Authentication now uses a browser-based OAuth authorization-code flow with PKCE. MCP clients open TrailBase’s existing login UI, and only current administrators are authorized. MCP access tokens are scoped and audience-bound to the instance’s `/mcp` URL.

That's cool and very interesting to me. Naively I was expecting folks would trust their agents and give them the credentials to do the authentication on their behalf but arguably this would be nicer to hand the agent tokens only. Will have to look a bit more into how it's wired up.

> MCP operations are routed through TrailBase’s existing in-process admin handlers to reduce drift and provide the administrative functionality discussed in the review, including schema changes, tables, indexes, triggers, Record APIs, configuration, users, files, backups, jobs, OAuth providers, and WASM components. Focused tools are also included for SQL, schema discovery, and configuration.

I'm a bit confused. Originally, i thought MCP endpoints would act as actuators (i.e. mediate the action) but then based on your original proposal I think I learned that MCP servers only share metadata, which is consumed and then used to talk to the actual endpoints (which in hindsight makes sense). Maybe you can do either, maybe you could shed some light on what the flow is, i.e. after calling the MCP and learning about the "list tables" tool, which endpoint does it call.

At the end of the day, I'm wondering if OpenAPI data would be enough thus allowing us to get rid of any tool duplication?

> I tested the complete flow with a disposable local TrailBase instance and RustRover via `mcp-remote`: browser login, PKCE callback, token exchange, MCP initialization, tool discovery, table creation, row insertion, and reads all succeeded.

I'll definitely will need to look into how to run and validate this myself in order to squelch my confusion and be able to maintain this.

FTR: I hadn't heard back so I've started some work around making OpenAPI better integrated and more complete hoping that this would also trivialize an MCP integration.

Independently, I was wondering about the desired execution model, which will depend on whether the MCP only serves metadata or actual data. If it's the former, it may make sense to colocate in the same binary but not necessarily in the same process. Specifically, should it be:

```
# Run MCP in the same proces
trail run --mcp
```

or

```
trail mcp
```

i.e. run a second process that serves the metadata. I don't have experience, would be interesting to hear what others do.

Thanks so much for this work.

brigon-dev:
Thanks for taking another look — these are good questions, and I think the main source of confusion is that MCP supports both descriptive and executable primitives.
An MCP server can expose resources/prompts that are primarily contextual metadata, but MCP tools are actuators. A client first calls tools/list to learn each tool’s name and JSON input schema. When the model uses one, the client sends a tools/call JSON-RPC request to /mcp; the MCP server executes the operation and returns its result.
In the current implementation, the flow for list_tables is:
IDE/agent
```
  -> POST /mcp with tools/call("list_tables")
  -> TrailBaseMcp::list_tables()
  -> existing admin router GET /tables
  -> existing list_tables_handler
  -> result returned through MCP
```
The admin request is dispatched through the in-process Axum router using the same AppState; it does not make a second HTTP request. execute_sql similarly dispatches to the existing POST /query admin handler. The generic call_admin_api tool accepts an HTTP method, admin-relative path, and body, then invokes the matching existing admin handler in-process.
So the MCP server is not only advertising endpoint metadata. It mediates and executes the actions.
I agree that OpenAPI should ideally be the source of truth. OpenAPI and MCP solve slightly different parts of the problem: OpenAPI can describe the HTTP operations and their schemas, while an MCP client still needs MCP tools—or a generic bridge tool—to invoke them. If the admin OpenAPI coverage becomes complete, I see a few possible ways to reduce duplication:

1. Keep one generic call_admin_api actuator and expose the OpenAPI document as context.
2. Generate MCP tool definitions from the admin OpenAPI document.
3. Keep only a small number of purpose-built tools where MCP needs behavior beyond a normal HTTP call, such as preserving/redacting configuration secrets.

I would be happy to align this PR with the OpenAPI work. The focused tools currently contain very little independent business logic: most are convenience wrappers around existing admin handlers. The old hand-maintained operation catalog has been removed.
On authentication: browser OAuth was intentional so the administrator’s password is entered only into TrailBase’s existing login UI. The agent receives a short-lived token scoped and audience-bound to that TrailBase /mcp resource. TrailBase also checks current administrator status from the database on every MCP request. A trusted client can still supply an existing admin bearer token as a compatibility option, but there is deliberately no MCP tool that asks the agent to handle an administrator password.
Regarding the execution model, I currently prefer:

`trail run --mcp`

because the MCP tools execute real administrative operations and can reuse the existing router, application state, authorization, configuration, and schema-refresh behavior directly. A separate:

`trail mcp`

process would either need to call the running TrailBase server over HTTP and manage separate administrative credentials, or open the same depot concurrently. The former starts to recreate the sidecar architecture, while the latter seems undesirable for database ownership and consistency.
If MCP ultimately becomes only an OpenAPI/metadata adapter, a separate process would make more sense. With the current actuator model, colocating it in the running process seems simpler and less error-prone. It remains disabled unless --mcp is explicitly supplied.
To validate locally:

```
trail --depot /tmp/trailbase-mcp components add trailbase/auth_ui

trail \
  --depot /tmp/trailbase-mcp \
  --public-url http://localhost:4000 \
  run \
  --address 0.0.0.0:4000 \
  --mcp
Then a command-based MCP client can connect with:
{
  "mcpServers": {
    "trailbase-local": {
      "command": "npx",
      "args": [
        "-y",
        "mcp-remote",
        "http://localhost:4000/mcp",
        "--allow-http",
        "--static-oauth-client-metadata",
        "{\"scope\":\"mcp\"}"
      ]
    }
  }
}
```

The focused tests can also be run with:

`cargo test -p trailbase --lib mcp::tests`

I’m very open to reducing or generating the tool surface once I understand the direction of the improved admin OpenAPI integration. In particular, I’d appreciate your preference between a generic OpenAPI-backed actuator and individually generated MCP tools.

ignatz:
Thanks for taking the time to discuss this with me - very much appreciated :pray: - and I'm sorry I'm not already more up to speed.

> I think the main source of confusion is that MCP supports both descriptive and executable primitives. An MCP server can expose resources/prompts that are primarily contextual metadata, but MCP tools are actuators. A client first calls tools/list to learn each tool’s name and JSON input schema. When the model uses one, the client sends a tools/call JSON-RPC request to /mcp; the MCP server executes the operation and returns its result. In the current implementation, the flow for list_tables is: IDE/agent
> 
> ```
>   -> POST /mcp with tools/call("list_tables")
>   -> TrailBaseMcp::list_tables()
>   -> existing admin router GET /tables
>   -> existing list_tables_handler
>   -> result returned through MCP

You're saying the agent calls the MCP and the MCP proxies the action.

My confusion also stems from browsing the landscape and stumbling over the code examples on https://crates.io/crates/rmcp-openapi, which seemed rather declarative but looking closer it does exactly what you said: "rmcp-openapi acts as a proxy between MCP clients and OpenAPI services..." :pray: 

> because the MCP tools execute real administrative operations and can reuse the existing router, application state, authorization, configuration, and schema-refresh behavior directly. A separate:
> 
> `trail mcp`
> 
> process would either need to call the running TrailBase server over HTTP and manage separate administrative credentials, or open the same depot concurrently. The former starts to recreate the sidecar architecture, while the latter seems undesirable for database ownership and consistency. If MCP ultimately becomes only an OpenAPI/metadata adapter, a separate process would make more sense. With the current actuator model, colocating it in the running process seems simpler and less error-prone. It remains disabled unless --mcp is explicitly supplied. To validate locally:

With my newfound understanding that the MCP actuates the action, this makes sense. Let me give the code a more proper look :pray:

mcp/README.md
```diff
@@ -0,0 +1,304 @@
...
+```
+
+`localhost` can use any port on which TrailBase is exposed, for example
+`http://localhost:4100/mcp`. `--allow-http` is required for a plain HTTP URL and
+is intended only for local development.
+
+### Public HTTPS URL
+
+Use this for Cloudflare Tunnel, a reverse proxy, or another deployed TrailBase
+instance:
```
ignatz, line 78: For my understanding, should an MCP ever be running in production? Naively, I would expected: no

mcp/README.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Development validation
+
+Run the native MCP unit tests with:
+
+```sh
+cargo test -p trailbase --lib mcp::tests
+```
+
+For an isolated manual test:
```
ignatz, line 296: ? - Isn't this just the dev setup, i.e. the way you'd normally run your mcp

ignatz:
I squashed and merged all the latest changes into a single commit: https://github.com/trailbaseio/trailbase/pull/276.

I'll use this as a base for the review, there's just less auxiliary changes. I'm happy to discuss here or there. Tentatively here to preserve the discussion but feel free to comment on either.

ignatz:
If we'd focus on dev-only use-cases (at least for now), and looking at mcp-remote:

> So far, the majority of MCP servers in the wild are installed locally, using the stdio transport. This has some benefits: both the client and the server can implicitly trust each other as the user has granted them both permission to run.

makes me wonder if `trail mcp` may make sense (on the side, I'm not too concerned about to processes sharing the depot, it may be a nuisance when the config is changed, we'd have to watch the file or SIGHUP the server). Presumably, not needing mcp-remote would at least make initial setup easier (something I'm still lacking :hide:)

(I'm also just liberally collecting my thoughts here as I'm reading up on the individual aspects)

ignatz:
Had some more time to meditate over it today.

I was contemplating what the use-cases for remote MCP are. Specifically, if a local mcp would run separately, you could still point it at a remote instance. Not that I'm advocating for unleashing agents on production instances but wouldn't splitting solve the issue? And by not actuating the action but merely proxying to the primary server there would also be no inconsistencies or multiple instances accessing the same depot.

I got a lot of sun, so chances are my brain got fried or did it?

As a strawman:

```sh
$ trail mcp --instance=prod.example.com --admin=admin@example.com
> password: ***
```

and then the thing would communicate locally via stdin/out

EDIT: I guess it wouldn't work for oauth admins. For local instances, given access to the depot, we wouldn't need sign-in at all

ignatz:
(sorry for sharing my ignorance and incremental discovery in such unstructured way, I partly want to document it for myself but even more so want to keep you in the loop especially if you have any insights to share).

In the (still) absence of a local agent-mcp-client, I discovered https://github.com/modelcontextprotocol/inspector, which has helped me to actually exercise the code (memo to myself: debugging oauth issues the client has unfortunate caching and needs wipage of ~/.mcp-inspector).

Overall the oauth works :tada: . However, how we construct the resource uris doesn't work for local use if a "site_url" is present. The connecting client will have a mismatch and refuse to proceed. If I fix that up, it works.

## A few more observations:

I get the tool list:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "tools": [
      {
        "name": "call_admin_api",
        "description": "Call a TrailBase admin API in-process. Paths are relative to /api/_admin. This exposes the same table, index, row, config, schema, query, user, log, backup, job, and WASM operations as the admin dashboard.",
        "inputSchema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema",
          "properties": {
            "body": {
              "default": null,
              "description": "Optional JSON request body."
            },
            "method": {
              "description": "HTTP method accepted by the TrailBase admin API.",
              "type": "string"
            },
            "path": {
              "description": "Admin API path relative to /api/_admin, including an optional query string.",
              "type": "string"
            }
          },
          "required": [
            "method",
            "path"
          ],
          "type": "object"
        },
        "outputSchema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema"
        }
      },
      {
        "name": "execute_sql",
        "description": "Execute SQL using TrailBase's admin query handler. Supports reads and writes; schema changes refresh cached metadata.",
        "inputSchema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema",
          "properties": {
            "attached_databases": {
              "default": null,
              "description": "Optional configured attached database names.",
              "items": {
                "type": "string"
              },
              "type": [
                "array",
                "null"
              ]
            },
            "query": {
              "description": "One or more SQLite statements. Schema-changing statements refresh TrailBase metadata.",
              "type": "string"
            }
          },
          "required": [
            "query"
          ],
          "type": "object"
        },
        "outputSchema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema"
        }
      },
      {
        "name": "get_config",
        "description": "Get the complete TrailBase configuration as protobuf text. Secret values are redacted.",
        "inputSchema": {
          "properties": {},
          "type": "object"
        }
      },
      {
        "name": "list_tables",
        "description": "List TrailBase tables, views, columns, indexes, triggers, and metadata.",
        "inputSchema": {
          "properties": {},
          "type": "object"
        },
        "outputSchema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema"
        }
      },
      {
        "name": "update_config",
        "description": "Validate and replace the TrailBase configuration using protobuf text from get_config. Existing secret values are preserved.",
        "inputSchema": {
          "$schema": "https://json-schema.org/draft/2020-12/schema",
          "properties": {
            "config": {
              "description": "Complete TrailBase config in protobuf text format, as returned by get_config.",
              "type": "string"
            }
          },
          "required": [
            "config"
          ],
          "type": "object"
        }
      }
    ]
  }
}
```

For some reason that I don't understand the inspector barfs: 

```
Rejected by the Inspector
Invalid result for tools/list: [ { "code": "invalid_value", "values": [ "object" ], "path": [ "tools", 0, "outputSchema", "type" ], "message": "Invalid input" }, { "code": "invalid_value", "values": [ "object" ], "path": [ "tools", 1, "outputSchema", "type" ], "message": "Invalid input" }, { "code": "invalid_value", "values": [ "object" ], "path": [ "tools", 3, "outputSchema", "type" ], "message": "Invalid input" } ]
```

Thought the tool list response overall looks good to me. However, it raises the next point...

... Because the tool definitions and APIs are separate, many actions would have to fall back on `call_admin_api`, which doesn't have a schema, which makes me wonder how an agent would reason about it. Is the expectation that it would reason about the entire code-base and figure out available APIs, IO schemas, ... by itself?

Lastly (and trivial to change), with the short-lived mcp-specific auth tokens and no means to refresh, is your expectation that users would re-auth every 10mins?

ignatz:
Just to summarize the options I see at the moment (please feel free to add):

1. As proposed: mcp-over-http co-located in the primary server process, custom oauth with custom auth/tokens, and custom tool definitions
    1. We could try to update the implementation to get the tools from OpenApi.
2. Use rmcp-openapi, which requires HTTP proxying, i.e. all interactions would be carried out by the primary server. In which case we need valid admin (auth,refresh,csrf) tokens. To sign-in we could:
    1. Use oauth as above but minting "standard" tokens rather than custom tokens. This would require MCP-over-http.
    2. If we wanted to use MCP-over-stdin:
        1. Login via CLI - would only work for non-oauth-admin users
        2. Privileged auto-login requiring access to depot (only at startup to get user and keys to mint tokens. Notably, this would require two arguments, server address and depot path. For a dev setup we could probably default to localhost:4000 :woman_shrugging: ).
        3. We could also just require the user to provide tokens, e.g. make them easy C&P from admin UI.
4. Standalone MCP (likely using stdin since everything would have to be local to depot anyway) that doesn't require a primary server to be up. I.e. point it at the depot and `oneshot` requests. Ideally tool definitions would still come from OpenApi and we'd have to convert ourselves.

Independent of the specific option, I think it would be wise to use OpenApi schemas that are already available to make the MCP tools more complete and useful.

Then there's the open question if we want to allow the MCP to connect to a remote TrailBase server, in which case only the options (2.i) and (2.II.a) would work, with (2.i) being the most general.

If remote access is not required (i.e. local-dev only), (3.) would be the most convenient to get started followed by (2.ii.b).

With my questions:
 * Do we want to support non-local workflows, i.e. run MCP-over-http in-process or have an MCP connect to a remote server? 
 * Is MCP-over-http broadly used/supported by clients?

I'm tentatively flip-flopping between (2.i) and (3.) (maybe even 2.ii.b or 2.ii.c, which would be the quickest to implement for a PoC or MVP. How the tokens are acquired, e.g. via CLI login, CLI arg or hijacking the depot is sort of a detail which "only" affects convenience). I don't see any clear advantage of (1.)  over (2.1), if one could simply connect an MCP server to an already running server as opposed to deployinng the server with MCP always enabled. Would love to hear your thoughts.

ignatz:
I have a basic implementation of 2.ii.c (very little code) here: https://github.com/trailbaseio/trailbase/blob/40957e62eaa59b6b6af3aea92581b9a720d9b383/crates/cli/src/bin/trail.rs#L490-L543

MCP inspector is very happy and finds a long list of structured tools:

Screenshot From 2026-08-14 17-44-07

I'm basically running:

```sh
TOKENS="eyJhdXR..." trail mcp --address=http://localhost:4000
```

Could this work for you?

EDIT: note that the PoC doesn't automatically refresh the auth-token yet.

ignatz:
After a bit more of ruminating and getting my hands dirty, I think the highest priority should be: complete, correct and comprehensive MCP tools, ultimately that will determine the quality of your experience. Details such as wire-transport and authentication are important, just not as important unless you have very specific use-cases.

My experiments are shaping up to the point where I have a fork of rmcp-openapi that strips out many dpeendencies and updates to current rmcp (yet the binary size increase is significant: ~7MB). I only support MCP over stdin but that seems to be what most local dev-flows and IDEs expect anyway. Authentication works either by providing tokens, e.g.:

```sh
trail mcp --tokens=$(trail --depot=client/testfixture user mint admin@localhost) http://mytrailbase.org
```

(or by looking the tokens up in the admin UI). Or with direct access to the depot:

```sh
trail --depot=client/testfixture mcp --user=admin@localhost http://mytrailbase.org
```

I hope that's workable

PR #264, opened 2026-07-18
Fix theme toggle icon showing wrong state on initial load

zyrakq:
`createTheme()` seeds its signal from `currentTheme()` before `initializeTheme()` applies the resolved theme to `<html>`. The signal is meant to resync via a `MutationObserver`, but that observer attaches in `createTheme()`'s own `onMount`, which fires *after* the parent `App`'s `onMount` already toggled the `dark` class — so the observer misses the mutation and the icon stays stuck on the stale initial state (moon shown on dark startup).

**Fix:** force `setTheme(currentTheme())` right before attaching the observer, resyncing to actual DOM state at mount time regardless of onMount order.

**Testing:** verified manually with dark preference on load — icon now correct immediately; toggling still works.

### DIFF
--- a/crates/assets/js/admin/src/lib/theme.ts
+++ b/crates/assets/js/admin/src/lib/theme.ts
@@ -38,9 +38,14 @@ export function createTheme(): Accessor<ResolvedTheme> {
     });
   });
 
-  onMount(() =>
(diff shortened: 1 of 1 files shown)


ignatz:
Much appreciated  :pray:

PR #265, opened 2026-07-18
Add WASM module registry and admin panel extensibility

### LINKED ISSUE #252: Allow WASM modules to extend the TrailBase admin panel

zyrakq:
Closes https://github.com/trailbaseio/trailbase/issues/252.

**What this adds:**

- `GET /_/admin/wasm-modules` — new API endpoint listing all loaded modules with their metadata
- **WASM Modules page** in the admin at `/wasm-modules` — shows each module with icon, display name, and description
- **Manifest via `get-manifest`** — module metadata (`display_name`, `icon`, `description`, `config_path`) is read through the guest's `get-manifest` export, following the manifest-based init approach. Partial manifests are supported — missing fields fall back to sane defaults
- **Module settings page** at `/wasm-modules/:name` — when a module declares a `config_path`, the admin fetches the fragment and wraps it in a custom element with Shadow DOM, rather than injecting it via `innerHTML`
- **`is_admin` check** added to `wasm-runtime-guest` — since this feature opens the door to admin-panel extensions, module endpoints that serve management UI must be able to guard access. The project didn't previously need this in the guest crate, but shipping admin-facing modules without a built-in auth check would push that burden onto every extension author individually

**Style isolation:**

The settings fragment is wrapped in a Web Component with Shadow DOM on the admin side, isolating styles in both directions.

**Real-world example:**

To verify the manifest-based init end-to-end, [wcauth-tb-component](https://github.com/zyrakq/wcauth-tb-component) can be used as a test module — it implements `get-manifest` and exposes a settings page. Build it from source, place the resulting binary as wcauth.wasm in traildepot, and it should appear in the WASM Modules page with its manifest applied.


ignatz:
I started to look at this, in the beginning there were some ~150 files flagged as changed by the GitHub review UI, which was partly due to the selected base-path, staleness of the manifest branch, ... I didn't really find a good way to filter out the noise, so I rebased `manifest` and cherry-picked your first change into https://github.com/trailbaseio/trailbase/tree/wasm-modules-page-in-admin-panel, to work off from :pray:

zyrakq:
Sorry, wrong base — I opened this against main instead of manifest, which is why manifest's commits are showing up in the diff. Glad you still managed to get a clean starting point out of it.

zyrakq:
One option worth trying: closing and reopening the PR sometimes forces GitHub to recompute the diff against the new base. I don't have that available on my end, but figured it's worth mentioning in case you want to give it a shot.

ignatz:
This change went verbatim (but squashed) as a foundation into v0.32.0. Thanks :pray:

PR #268, opened 2026-07-30
fix(dart): robust SSE chunk parsing for multi-event TCP packets

devsdocs:
The original connectSse logic in lib/src/sse.dart used a BytesBuilder to buffer incoming TCP packets from the http.StreamedResponse. When it detected a chunk ending in \n\n, it blindly assumed that the entire buffer contained exactly one complete SSE event, which it passed to jsonDecode.

Why it crashes: In high-throughput real-time streams (like a bulk data sync), multiple SSE events can be concatenated and delivered in a single TCP chunk by the underlying network stack. If a packet arrives containing two events merged together (e.g., data: {...}\n\ndata: {...}\n\n), the BytesBuilder passes the whole string into jsonDecode. jsonDecode successfully parses the first object but crashes immediately upon hitting the trailing data: keyword of the second event, throwing FormatException: Unexpected character.

The Fix: The payload parsing was updated to use a String buffer that loops through the buffer using indexOf('\n\n'). It now recursively extracts and isolates every individual event from the stream chunk, ensuring jsonDecode is only ever handed a single, clean JSON string per parse.

### DIFF
--- a/client/dart/lib/src/sse.dart
+++ b/client/dart/lib/src/sse.dart
@@ -1,6 +1,5 @@
 import 'dart:async';
 import 'dart:convert';
-import 'dart:typed_data';
 
 import 'package:meta/meta.dart';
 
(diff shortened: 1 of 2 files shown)


ignatz:
Thanks for the fix. Much appreciated :pray: 

I pushed your change verbatim to the dev branch: https://github.com/trailbaseio/trailbase/commit/b4d1b25755e0e62ae525df9e5be124057911cb26. I did a small follow-up to delay the utf8 decoding (closer to the original implementation) and expect a release in the next ~hour.

:heart:

PR #270, opened 2026-08-03
fix: Python Client Calls Wrong API

UserNobody14:
I found a bug: Python's promote_anonymous posts to `/login_anonymous` instead of `/promote_anonymous` (client/python/trailbase/__init__.py:490). Unfortunately the tests weren't really checking for it, so it appears to have been failing silently. I added tests that will fail with the old url and only pass with the new one.


ignatz:
Much appreciated :pray: - Will merge ASAP and cut a new release.

Issue #215, opened 2026-02-23
OpenID Connect scopes unsupported


UnseenFaith:
The base open id connect provider assumes that the scopes email and profile are available via the external provider, but that isn't always the case. It would be nice to configure these scopes, or have the ability to specify the well known configuration and have trailbase pull the allowed scopes and claims via a fetch request to the url. The latter would be nice in general since you could use the configuration to fill out the auth/token/profile urls, but ultimately I think adjusting scopes manually would be more suited and the most customizable

EDIT: I realize this is probably due to how trailbase expects authentication to work across all providers/stored in data, so removing those two as it stands would probably break stuff.

UnseenFaith:
Nah it's not to avoid using emails, it's just a matter of the external provider (Twitch) doesn't actually provide those scopes. I was just messing around with the idea of trailbase and found out that it wasn't supported by the base implementation. I don't know of any other providers that do this, but twitch specifically uses scopes to determine what access a user has to certain APIs. I think technically you could get the email via claims? They have claim support via query parameters, again though haven't really interacted with another provider that does stuff like this so it might be more rare/personal.

UnseenFaith:
Alrighty I pulled the new version and have been testing it. I'm not sure if I'm doing something incorrect or if it's broken so I'll explain how I think the flow should work.

- User hits login on frontend which routes to trailbaseUrl/api/auth/v1/oauth/twitch/login with a redirect_uri back to the frontend
- Trailbase does it's shenanigans to forward the user to twitch, user approves and twitch verifies the callback, which would be the trailbase callback uri here
- Trailbase authenticates with the code returned in the url and verifies that everything is correct, and then should redirect back to the redirect uri provided initially in the login route call

Right now, after approving the application, it gets redirected to trailbase with an HTTP Method 424 error and seems like it's stuck between exchanging and redirecting the user back to the frontend?

Let me know if I'm misunderstanding or have set something up incorrectly here.

> It does sound a bit like hickerey :) - How would you typically get the intercepted token from your "edge midleware" running in your auth server to the backend? Genuinely curious

I've seen a couple of examples. Sometimes you can just edit directly what the database stores to tack on the tokens, other times you can attach the tokens somewhere (usually the profile object) just so you have access to them, and then store them in a separate table or however the developer wants.

From there you can just pull the token from the database and use it, or wherever it was stored.

ignatz:
[...]

PR #273, opened 2026-08-03
draft: Refactor OAuth Social Providers

UserNobody14:
Refactored the repetitive oauth provider code into a series of handy flexible traits, and added tests for the many different quirks of the providers.

I also believe this is a solution to https://github.com/trailbaseio/trailbase/issues/215, allowing more configurable scopes to be entered, and eliminating email requirements for providers that may not furnish them for one reason or another.

Looking at ways to reduce the line count on this one, i still feel its a bit repetitive.


ignatz:
Had a first skim. Looks promising. Never loved the verbosity, so pulling in a second layer for providers that can remain mostly declarative makes a lot of sense :+1: 

My only concern with the OAuth code in general is that it's hard to impossible to test :/. Did you have the chance to test a one or more providers? :pray: 

EDIT: Are you still iterating or is this good to go?

UserNobody14:
Sorry, wound up caught up in some personal stuff. this is good to go as far as tests and all, but idk, i still feel like its not quite right. Maybe just perfectionism

ignatz:
FTR: I started upstreaming the most critical changes proposed here. There's a lot going on and I have a hard time wrapping my head around all of them at once (me problem) while ensuring the behavior doesn't change :)

What happened:

* I first tidied up the base trait. The plan was to improve the baseline and in principle allow for golden tests to make sure the behavior doesn't change (though I was to lazy to write tests for all providers).
* Then I pulled out the configurable scopes proposed here. I made it so that OAuthUser can handle optional emails (I see the intent, maybe got lost in the chain?) and policy enforcement.
* Also added the ability for Social OAuth provider to scope down based on UserIdentifier policy.
* Added your Admin UI changes

:pray: 

IUC, the only thing still missing (probably the lowest priority) is to make "simple" Social providers less verbose by having a shared parametrized implementation. I'll probably also look into this but in due time.

ignatz:
Update: I simplified many of the providers very much inspired by your proposal incorporating all the other changes (conditional scopes, simplified base trait). It's not quite declarative and I'm sure lines could be saved but it's pretty thin on abstractions and w/ scopes being more conditional, this feels like a reasonable trade-off: https://github.com/trailbaseio/trailbase/blob/dev/crates/core/src/auth/oauth/providers/gitlab.rs

This was some of the most finicky code, thanks for stepping on my toes and this awesome PR :pray:

PR #276, opened 2026-08-12
Add native MCP server to TrailBase

ignatz:
Squash rebase of the latest (native) changes of https://github.com/trailbaseio/trailbase/pull/262/changes#diff-3009e2cec212fff3b4c6741f36eba2016178db426fd526cd3fbd861a9f6236b5

Mostly just wanted the starting point for review.


brigon-dev:
Thanks for landing `trail mcp`. The OpenAPI-generated tool list solves the exact drift problem I couldn't with a hand-written catalog, and stdio-only is the right call for how most IDEs actually connect.

One gap I'd like to propose closing, additively, without touching the stdio transport or tool generation. Right now getting a token into `trail mcp` means either `--user` plus local depot access, or manually copying a token out of the admin UI's profile dialog into `TOKENS=`. For a remote server that's the one piece of friction left. Most hosted MCPs (Stripe, AWS, etc.) handle auth so the browser pops up once and everything after is automatic.

Proposal: add a login mode to `trail mcp` using the loopback pattern `gh auth login` / `gcloud auth login` / `aws sso login` use:

1. `trail mcp` generates a PKCE pair and starts a short-lived HTTP listener on `127.0.0.1:<random port>`.
2. It opens the user's browser to TrailBase's existing `AuthorizationCodeFlowWithPkce` login page (already used for browser app clients) with that loopback address as `redirect_uri`.
3. User logs in on TrailBase's real login page. `trail mcp` never sees a password.
4. TrailBase redirects to the loopback listener with the code. `trail mcp` exchanges it for tokens directly and caches them locally, refreshing silently the same way `--user` mode already does.

The one server-side change this needs: `validate_redirect_impl` in `crates/core/src/auth/util.rs` already special-cases loopback hosts (127.0.0.1/::1/localhost, any port), but only when `state.dev_mode()` is true. I'd propose that bypass apply unconditionally, which is the standard justification behind RFC 8252's native-app redirect handling. Anything that can bind to `localhost` on the user's machine already has local code execution, so trusting a loopback redirect there doesn't add new attack surface, even in prod.

Nothing here touches `rmcp-openapi`, the tool schema, or the stdio transport. It's purely a different way for `trail mcp` to acquire its token, opt-in alongside `--user` and `TOKENS`. Happy to put up a draft PR if this direction looks reasonable to you first.

ignatz:
> Thanks for landing `trail mcp`. The OpenAPI-generated tool list solves the exact drift problem I couldn't with a hand-written catalog, and stdio-only is the right call for how most IDEs actually connect.
> 
> One gap I'd like to propose closing, additively, without touching the stdio transport or tool generation. Right now getting a token into `trail mcp` means either `--user` plus local depot access, or manually copying a token out of the admin UI's profile dialog into `TOKENS=`. For a remote server that's the one piece of friction left. Most hosted MCPs (Stripe, AWS, etc.) handle auth so the browser pops up once and everything after is automatic.
> 
> Proposal: add a login mode to `trail mcp` using the loopback pattern `gh auth login` / `gcloud auth login` / `aws sso login` use:
> 
>     1. `trail mcp` generates a PKCE pair and starts a short-lived HTTP listener on `127.0.0.1:<random port>`.
> 
>     2. It opens the user's browser to TrailBase's existing `AuthorizationCodeFlowWithPkce` login page (already used for browser app clients) with that loopback address as `redirect_uri`.
> 
>     3. User logs in on TrailBase's real login page. `trail mcp` never sees a password.
> 
>     4. TrailBase redirects to the loopback listener with the code. `trail mcp` exchanges it for tokens directly and caches them locally, refreshing silently the same way `--user` mode already does.
> 
> 
> The one server-side change this needs: `validate_redirect_impl` in `crates/core/src/auth/util.rs` already special-cases loopback hosts (127.0.0.1/::1/localhost, any port), but only when `state.dev_mode()` is true. I'd propose that bypass apply unconditionally, which is the standard justification behind RFC 8252's native-app redirect handling. Anything that can bind to `localhost` on the user's machine already has local code execution, so trusting a loopback redirect there doesn't add new attack surface, even in prod.
> 
> Nothing here touches `rmcp-openapi`, the tool schema, or the stdio transport. It's purely a different way for `trail mcp` to acquire its token, opt-in alongside `--user` and `TOKENS`. Happy to put up a draft PR if this direction looks reasonable to you first.

Thanks for the feedback. I'll have a look. I'm not familiar with the loopback pattern (yet)

PR #286, opened 2026-09-10
native apple signin

yurvon-screamo:
Native Sign in with Apple login endpoint

Adds POST /auth/v1/oauth/apple/native for native clients (iOS/macOS apps using ASAuthorizationController), complementing the existing web OAuth flow:

- RS256 verification against Apple's JWKS by kid, with configurable audience: the native flow binds tokens to the App ID, distinct from the web flow's Services ID client_id (new optional native_client_id proto field, field 16).
- Anti-replay nonce claim: client sends the SHA-256 hash of its raw nonce with the request (Firebase pattern); the endpoint re-hashes and compares. Malformed tokens/nonce mismatches are rejected 400 before any JWKS fetch (no outbound calls for garbage).
- User matching by team-stable Apple sub through the same provider-id + provider-user-id path as the web callback — web-created and native accounts resolve to the same user. create_user_for_external_provider extracted into shared oauth/users.rs for both flows.
- Fail-closed: unconfigured native_client_id → 500, signature/aud/iss/exp failures → 401, first login without verified email → 424.
- Full endpoint-level test coverage with an injected JWKS fixture (RSA keypair, kid lookup, audience/nonce/expiry negatives).


crates/core/proto/config.proto
```diff
@@ -75,6 +75,14 @@ message OAuthProviderConfig {
...
   /// username-based `UserIdentifier`.
   repeated string scopes = 15;
+
+  /// Apple only: expected `aud` of identity tokens issued by the native
+  /// Sign in with Apple flow (ASAuthorizationController). Native tokens are
+  /// audience-bound to the App ID (bundle identifier), while the web OAuth
+  /// flow validates against `client_id` (the Services ID). The two audiences
+  /// differ, so the native one needs its own config entry. When unset, the
+  /// native login endpoint fails closed.
+  optional string native_client_id = 16;
```
ignatz, line 85: Disclaimer: I will have to read up more on apple's proprietary native auth flow.

I don't think we should conflate native_client_id with OAuth. It's not used in the oauth flow. I can understand the desire to shoehorn it into the existing infra to benefit from the config UI, etc. In that case, it may make sense to model it as a separate "OAuth-like" provider, e.g. apple-native besides apple-oauth (in which case client_id would suffice).

Alternatively, we could or maybe should separate config and implementation entirely. After all they are entirely separate flows. WDYT?

crates/core/src/auth/oauth/providers/apple.rs
```diff
@@ -97,34 +105,128 @@ impl AppleOAuthProvider {
     http_client: &reqwest::Client,
     id_token: &str,
   ) -> Result<AppleIdToken, AuthError> {
-    let header = jsonwebtoken::decode_header(id_token)
-      .map_err(|err| AuthError::FailedDependency(err.into()))?;
-    let Some(kid) = header.kid else {
+    // Header first: malformed tokens are rejected before the JWKS fetch
+    // (outbound calls only for structurally valid tokens).
+    let header = jsonwebtoken::decode_header(id_token).map_err(|err| {
+      log::warn!("Apple id_token header could not be decoded: {err}");
```
ignatz: I'd be careful with all the logging. It's also something I  removed from prior PRs. The criticality is extremely context dependent and the cause is non-local (this sounds more like an apple problem), which would spam your local logs. I don't think it's inherently wrong to log. I only think that high criticality should be left to local and inherently critical issues.

crates/core/src/auth/oauth/apple_native.rs
```diff
@@ -0,0 +1,492 @@
...
+  let oauth_user = OAuthUser {
+    provider_user_id: claims.sub,
+    provider_id: proto::OAuthProviderId::Apple,
+    email: claims.email,
+    username: None,
+    verified: claims.email_verified.is_some_and(|v| v.value()),
+    avatar: None,
+  };
+
+  let db_user = get_or_create_native_user(&state, oauth_user).await?;
```
ignatz: This may be a bit misleading since there's nothing OAuthy about it. The get_or_create_native_user impl is mostly just validating the inputs, which seems unecessary if you just passed the expected inputs directly

crates/core/proto/config.proto
yurvon-screamo: Agreed - native_client_id doesn't belong in OAuthProviderConfig. The native flow isn't OAuth at all: no code exchange, no client secret, no redirect - the app gets the id_token itself via ASAuthorizationController, the server just verifies it (JWKS + aud + nonce).

One constraint: both flows share one identity space. Apple's sub is stable per team, so web (Services ID) and native (App ID) logins of the same user must resolve to the same account - user matching and token minting should stay shared.

So I'd go with your second option:
- separate config entry, e.g. apple_native_app_id on AuthConfig - that's all the flow needs;
- move the endpoint out of oauth/ (only reuses the JWKS helpers + user lookup);
- drop native_client_id from OAuthProviderConfig.

The second-provider variant seems risky: own provider id - two accounts for one person; shared provider id - ambiguous which entry drives the web flow.

Minimal Apple-only field, or a generic identity_token_providers section (Google native would fit later)? 

Happy to rework the PR - small diff either way.

crates/core/src/auth/oauth/providers/apple.rs
yurvon-screamo: Agreed, I'll make the correction.

crates/core/src/auth/oauth/apple_native.rs
yurvon-screamo: Agreed, I'll make the correction.

crates/core/proto/config.proto
ignatz: > The second-provider variant seems risky: own provider id - two accounts for one person; shared provider id - ambiguous which entry drives the web flow.

Good point (every TB user is tied to a specific provider)

> Happy to rework the PR - small diff either way.

Moving out of OAuth sounds good :pray:

ignatz: BTW, it's fine if the new setting isn't yet represented in the UI or we do it later/separately (or leave it w/o a specific plan, IMHO it's ok for the UI to represent a subset).

ignatz: Thanks for pulling out the config :pray: . I've squashed and taken your change a little further, splitting native_apple further from oauth: https://github.com/trailbaseio/trailbase/pull/288 .

ignatz: I'm starting to be decently happy with where the code's at. Would like to have a look before moving forward?

PR #287, opened 2026-09-11
Fix creating records using default values with Kotlin client

Bnyro:
Hey!

By default, kotlinx serialization doesn't serialize a field if it's the same as the default value.

E.g.
```kt
@Serializable
data class Foo(val x: Int = 3)

Json.encodeToString(Foo())
```
yields an empty dict `{}`.

This cause the issue that when updating a record with a default value, the value does not get sent to the server and we get `db constraint: not null`.

Forcing kotlinx serialization to send default values too this fixes it.

### DIFF
--- a/client/kotlin/lib/src/commonMain/kotlin/Library.kt
+++ b/client/kotlin/lib/src/commonMain/kotlin/Library.kt
@@ -751,6 +751,7 @@ private fun initClient(): HttpClient {
       json(
               Json {
                 ignoreUnknownKeys = true
+                encodeDefaults = true
                 isLenient = true
               }
       )


ignatz:
Much appreciated :pray: - will merge and release a new client ASAP

ignatz:
I did modify the change slightly to cover all serialization consistently. New client version is released :pray:

Bnyro:
Thank you! I missed that there was a second `Json` instance.

Great that you already pushed a release so quick, then I can continue to use the released version from Gradle instead of my fork which makes it much easier :)

Bnyro:
A note for anyone who comes across the issue that when creating an entry, the backend returns `db constraint: not null`. That's probably because we now send `id = null`, and trailbase then rejects the request instead of auto-generating the id.

One solution is to set `@EncodeDefault(EncodeDefault.Mode.NEVER)`, which forces Kotlin Serialization to only send the value of `id` if it's not the default (i.e. null).
E.g.
```kotlin
@Serializable
data class MyRecordType(
    @EncodeDefault(EncodeDefault.Mode.NEVER)
    val id: String? = null,
    val name: String,
    ...
)
```

Possibly trailbase could, when it gets a null value as primary key, auto-generate a value for it (as it currently does when the value is not set), but I'm not sure if that would break other stuff.

To clarify: Sending
```json
{}
```
works and auto-generates the primary key, but
```json
{"id": null}
```
doesn't.

ignatz:
I feel a bit bedazzled now. I'm not too versed with kotlin, is this a result of the explicit default, i.e.

```kotlin
@Serializable
data class MyRecordType(
    val id: String? = null,
    ...
)
```

? What if you remove the `= null` would it still get serialized as `{ "id": null }`? In which case I'll revert the change.

Note that the requirements are derived from the database schema and null is different from default, e.g. you may have a table:

```sql
CREATE TABLE foo (value STRING DEFAULT('test'));
```

which is a nullable column with a non-null default. Abesence in the JSON should set it to `test`, `{ "value": null }` should set it to null.

If the presence of explicit default values affects the inferred serialization code, it's the responsibility of the code to get it right or use a code-generator to generate different CREATE/UPDATE/READ types from the JSON schema. Using non-null default values for a data-class seems problematic because defaults may go out of sync with database defaults :woman_shrugging:

ignatz:
To answer my own question, the explicit default has nothing to do with the serialization, thus I'll revert this PR. Your types should reflect the respective requirements insert, read, update, which may all be different, e.g. default don't apply on update.

Bnyro:
You're right, I didn't think about fixing it by making the database and the client models have the same default values. That's a cleaner fix.

A lot of confusion could probably be avoided by having a README for the Kotlin client library or some other sort of documentation that mentions how the models are supposed to look and how `@Serializable` is supposed to be used, the requirements on default values, etc....

I also had to search for a while until I figured out where the client library is published (i.e. `io.trailbase:trailbase-client`) because the documentation on trailbase.io only points to https://github.com/trailbaseio/trailbase/tree/main/client/kotlin which doesn't have any reference to that. (just found that the https://mvnrepository.com/artifact/io.trailbase/trailbase-client is linked in the README, but I expected to somehow find that in the docs or the kotlin library README)

ignatz:
Thanks for all the feedback :pray: 

> You're right, I didn't think about fixing it by making the database and the client models have the same default values. That's a cleaner fix.

TBH, I'm not sure. They may diverge and the only true source of truth will be your database schema. To me it's a bit of an odd default that kotlin's serializer may skip non-null in-code default values (I've never seen this in any other language, arguably this is a library and not a language idiosyncrasy). Aforementioned skew problem is true for any persistent storage, e.g. you may write JSON to disk yesterday and read it today with a different version of your program leading to a re-interpretation of yesterday's writes :woman_shrugging: 

Anyway, I've been further tuning the JSON serializer setup and I think I found the appropriate setting where non-null default values are serialized and nulls are skipped.

In any case, I really appreciate you bringing this up, I had no idea (arguably a me problem).

> A lot of confusion could probably be avoided by having a README for the Kotlin client library or some other sort of documentation that mentions how the models are supposed to look and how @Serializable is supposed to be used, the requirements on default values, etc....

Agreed. FWIW, as a human this feedback makes me very happy that folks are still reading docs. As a human, I dread this feedback because having to write documentation for 8 clients at a quality-level that makes it worth reading is a huge investment.

We should absolutely add and improve the lackluster documentation. Thus far, I've tried to stay under the radar by keeping the clients simple, intuitive and as consistent as possible. I hope the latest serialization tweak restores some of that intuitive consistency :hide:

> I also had to search for a while until I figured out where the client library is published (i.e. io.trailbase:trailbase-client) because the documentation on trailbase.io only points to https://github.com/trailbaseio/trailbase/tree/main/client/kotlin which doesn't have any reference to that. (just found that the https://mvnrepository.com/artifact/io.trailbase/trailbase-client is linked in the README, but I expected to somehow find that in the docs or the kotlin library README)

Yikes, it was in the README.md but not the website - fixed :pray: 

Looking at the links, I have a silly, ignorant question. I haven't really changed anything about my publishing practices but looking at https://mvnrepository.com/artifact/io.trailbase/trailbase-client, the library sits at 0.7.1, while looking at https://central.sonatype.com/artifact/io.trailbase/trailbase-client it sits at 0.7.5. Am I holding it wrong or was there some kind of ecosystem change?

Bnyro:
> TBH, I'm not sure. They may diverge and the only true source of truth will be your database schema. To me it's a bit of an odd default that kotlin's serializer may skip non-null in-code default values (I've never seen this in any other language, arguably this is a library and not a language idiosyncrasy). Aforementioned skew problem is true for any persistent storage, e.g. you may write JSON to disk yesterday and read it today with a different version of your program leading to a re-interpretation of yesterday's writes 🤷‍♀️
> 
> Anyway, I've been further tuning the JSON serializer setup and I think I found the appropriate setting where non-null default values are serialized and nulls are skipped.
> 
> In any case, I really appreciate you bringing this up, I had no idea (arguably a me problem).
>
I looked at your changes, thanks for re-investigating! I think that's a solution everybody can be happy with, but I'll test it further in my trailbase-based project in the coming days :)

The only problem I see is that if you now send value = null` (with a create or update operation) `, while the database has `value = 3` as default. Because we skip null values, `value = null` doesn't get sent. Hence, the database uses the default value of 3, although the user sent `null`. In 99% of the cases, the default value is probably `null` in the database scheme, so this won't cause any issues, but in some cases people might decide to use a non-null value as default.

So we should probably document this case, and think about a solution how we can fix this on client-side. I searched for an annotation like  `@AlwaysSendExplicitNulls` or something, but that doesn't seem supported (see [kotlinx annotations source code](https://github.com/Kotlin/kotlinx.serialization/blob/master/core/commonMain/src/kotlinx/serialization/Annotations.kt)). So not sure what to do here other than politely asking people to use null as default if a value is nullable if they use the Kotlin client, which is also weird? Or maybe my conclusions here are somewhat wrong, maybe I'm missing something and everything works correct now out of the box here?

> Agreed. FWIW, as a human this feedback makes me very happy that folks are still reading docs. As a human, I dread this feedback because having to write documentation for 8 clients at a quality-level that makes it worth reading is a huge investment.
>
> We should absolutely add and improve the lackluster documentation. Thus far, I've tried to stay under the radar by keeping the clients simple, intuitive and as consistent as possible. I hope the latest serialization tweak restores some of that intuitive consistency :hide:
>
Agreed. Thanks for taking the time to work on this PR here :)

> Yikes, it was in the README.md but not the website - fixed 🙏
>
Much better, thanks! Perhaps it would be useful to write a short "getting started" guide on the Sonatype page (i.e. creating a client, logging in and accessing records) - I can do a draft that if you tell me where to put it. Perhaps it just takes the README.md of the kotlin library as its description?

> Looking at the links, I have a silly, ignorant question. I haven't really changed anything about my publishing practices but looking at https://mvnrepository.com/artifact/io.trailbase/trailbase-client, the library sits at 0.7.1, while looking at https://central.sonatype.com/artifact/io.trailbase/trailbase-client it sits at 0.7.5. Am I holding it wrong or was there some kind of ecosystem change?
>
https://repo1.maven.org/maven2/io/trailbase/trailbase-client/0.7.5/ has the latest version, so it's definitely published properly. I would assume that mvnrepository just takes some time to re-index the full Maven Central repo and display all new versions (but I'm not an expert on that).

ignatz:
> but I'll test it further in my trailbase-based project in the coming days :)

Appreciated :pray: 

> The only problem I see is that if you now send value = null`(with a create or update operation)`, while the database has `value = 3` as default. Because we skip null values, `value = null` doesn't get sent. Hence, the database uses the default value of 3, although the user sent `null`. In 99% of the cases, the default value is probably `null` in the database scheme, so this won't cause any issues, but in some cases people might decide to use a non-null value as default.

Yes for the create case. Arguably even worse for the update case, when one wants to unset a prior value. Both very valid :pray: . The server handles it correctly but the client doesn't provide the necessary per-field, per-use-case granularity.

I wanted to see what quicktype does and they seem to use a different JSON library (klaxon) all together:

Screenshot From 2026-09-14 11-41-45

Not sure if it's related but I very much agree with you that this should be addressed and documented.

> Much better, thanks! Perhaps it would be useful to write a short "getting started" guide on the Sonatype page (i.e. creating a client, logging in and accessing records) - I can do a draft that if you tell me where to put it. Perhaps it just takes the README.md of the kotlin library as its description?

Appreciated. I trust you, you seem a lot more familiar with the ecosystem.
 
> > Looking at the links, I have a silly, ignorant question. I haven't really changed anything about my publishing practices but looking at https://mvnrepository.com/artifact/io.trailbase/trailbase-client, the library sits at 0.7.1, while looking at https://central.sonatype.com/artifact/io.trailbase/trailbase-client it sits at 0.7.5. Am I holding it wrong or was there some kind of ecosystem change?
> 
> https://repo1.maven.org/maven2/io/trailbase/trailbase-client/0.7.5/ has the latest version, so it's definitely published properly. I would assume that mvnrepository just takes some time to re-index the full Maven Central repo and display all new versions (but I'm not an expert on that).

Thanks for having a look. If it takes a while, it certainly takes a while. Days have passed, so if the artifacts are properly published and accessible, I'll simply ignore mvnrepository from now one :woman_shrugging:

Bnyro:
> I wanted to see what quicktype does and they seem to use a different JSON library (klaxon) all together:
> Screenshot From 2026-09-14 11-41-45
> 
> Not sure if it's related but I very much agree with you that this should be addressed and documented.
> 
Kotlinx Serialization is (in my opinion) the best library for serialization in the Kotlin ecosystem. For example, a lot of Jetpack Compose standard library for creating Android/Kotlin multiplatform apps (e.g. compose navigation) depends on Kotlinx Serialization. So chances are high that users of the trailbase library already have Kotlinx serialization as a dependency anyways, so it doesn't really make sense to add an other library for serialization purposes. So I don't think it makes sense to switch to something else.

> Appreciated. I trust you, you seem a lot more familiar with the ecosystem.
>
I never published a Java/Kotlin library, so I'm not sure if I'm more familiar. But I'll give it a try :)

Bnyro:
> A note for anyone who comes across the issue that when creating an entry, the backend returns `db constraint: not null`. That's probably because we now send `id = null`, and trailbase then rejects the request instead of auto-generating the id.
> 
> One solution is to set `@EncodeDefault(EncodeDefault.Mode.NEVER)`, which forces Kotlin Serialization to only send the value of `id` if it's not the default (i.e. null). E.g.
> 
> ```kotlin
> @Serializable
> data class MyRecordType(
>     @EncodeDefault(EncodeDefault.Mode.NEVER)
>     val id: String? = null,
>     val name: String,
>     ...
> )
> ```
> 
Rethinking it, I think it might make sense to revert back to `explicitNulls = true` and add a note that the ID should be annotated with `@EncodeDefault(EncodeDefault.Mode.NEVER)` when creating/updating database records.

That approach isn't perfect, but it doesn't have the downside that it's impossible to reset values to `null`. I guess it's pretty common that an update statement is supposed to remove the value of a field that previously had a value by sending it as `null`. As we discussed, not sending it as `explicitNulls = false` does basically prevents you from ever doing that because TrailBase doesn't change the value if it's omitted from the request.

Or we revert to how it was before the PR and explicitly state that things will break if the Kotlin classes and database scheme's defaults are inconsistent.

ignatz:
> > A note for anyone who comes across the issue that when creating an entry, the backend returns `db constraint: not null`. That's probably because we now send `id = null`, and trailbase then rejects the request instead of auto-generating the id.
> > One solution is to set `@EncodeDefault(EncodeDefault.Mode.NEVER)`, which forces Kotlin Serialization to only send the value of `id` if it's not the default (i.e. null). E.g.
> > ```kotlin
> > @Serializable
> > data class MyRecordType(
> >     @EncodeDefault(EncodeDefault.Mode.NEVER)
> >     val id: String? = null,
> >     val name: String,
> >     ...
> > )
> > ```
> 
> Rethinking it, I think it might make sense to revert back to `explicitNulls = true` and add a note that the ID should be annotated with `@EncodeDefault(EncodeDefault.Mode.NEVER)` when creating/updating database records.
> 
> That approach isn't perfect, but it doesn't have the downside that it's impossible to reset values to `null`. I guess it's pretty common that an update statement is supposed to remove the value of a field that previously had a value by sending it as `null`. As we discussed, not sending it as `explicitNulls = false` does basically prevents you from ever doing that because TrailBase doesn't change the value if it's omitted from the request.
> 
> Or we revert to how it was before the PR and explicitly state that things will break if the Kotlin classes and database scheme's defaults are inconsistent.

I tend to agree with your assessment, as in: at least there is a path. I was just hoping to find a solution that wouldn't require a new type declaration per use-case. It would be very odd, if folks have 25 different `MyRecordType<N>` classes for different updates to `MyRecord`. I was looking for prior art and came across a few custom serializers like: https://github.com/Osmerion/Omittable

Alternatively for updates something more literal (if kotlin allows for it) but less type-safe may be nice:

```kotlin
api.update(id, {
  "col_to_unset": null
})
```

Bnyro:
> I tend to agree with your assessment, as in: at least there is a path. I was just hoping to find a solution that wouldn't require a new type declaration per use-case. It would be very odd, if folks have 25 different MyRecordType<N> classes for different updates to MyRecord.

> Alternatively for updates something more literal (if kotlin allows for it) but less type-safe may be nice:

> api.update(id, {
>   "col_to_unset": null
> })
>
In Kotlin, that would be 
```kotlin
api.update(id, mapOf("col_to_unset" to null))
```
, but that's not really an exciting solution either.

> I was looking for prior art and came across a few custom serializers like: https://github.com/Osmerion/Omittable
>
I think that the idea of the Omittable serializer looks quite nice, because it does exactly what we need. If we document it properly, it should be fairly easy to understand and use. Though it might be easier to implement our own Omittable type because this can be implemented in less than 30 lines and is probably not worth an extra dependency because it does a lot of things we don't need and isn't really Kotlin-ish (i.e. doesn't have a nice toOmittable extension function to do `myObject.toOmittable()` instead of `Omittable.of(myObject)` which is a Java and not a Kotlin pattern). At least that allows people to be 100% sure what's happening because it's explicitly shown and there's no black magic in what Kotlinx serialization does.

On the downside, this prevents you from auto-generating these classes from your database layout (because the automatic SQL to Kotlin class converters don't support it) and it adds a lot of additional complexity to construction objects (i.e. you have to use `val x = Foo(Omittable.of("foo"), Omittable.of(3))`) which can become quite annoying I guess.

Not sure what you prefer, I'm open for both. You probably know better how it works in the other client libraries for TrailBase, so maybe it makes sense to stay consistent to how they work.

ignatz:
> I think that the idea of the Omittable serializer looks quite nice, because it does exactly what we need. 

glad we have an avenue.

> On the downside, this prevents you from auto-generating these classes from your database layout (because the automatic SQL to Kotlin class converters don't support it) 

I'm not sure I follow. Are you driving code-generation by parsing a CREATE TABLE statement? I would recommend driving code-generation from JSON schemas. Right now TB, doesn't help with that and defers to other, idiomatic JSON schema to your language converters (like quicktype).

> and it adds a lot of additional complexity to construction objects (i.e. you have to use val x = Foo(Omittable.of("foo"), Omittable.of(3))) which can become quite annoying I guess.

Would that something we could address if we absorb the Omittable?

> Though it might be easier to implement our own Omittable type because this can be implemented in less than 30 lines and is probably not worth an extra dependency because it does a lot of things we don't need and isn't really Kotlin-ish (i.e. doesn't have a nice toOmittable extension function to do myObject.toOmittable() instead of Omittable.of(myObject) which is a Java and not a Kotlin pattern).

I'll defer to you as to what is idiomatic. I'm happy to absorb if it provides a canonical way to handling null vs absent.

> Not sure what you prefer, I'm open for both. You probably know better how it works in the other client libraries for TrailBase, so maybe it makes sense to stay consistent to how they work.

Generated bindings is still an area where many of the implementations fall short, the experience would probably be similar. Structurally typed languages like TypeScript (which is probably the most used client) don't have this problem. Some languages distinguish between null and undefined. Whatever we cook up here, will likely inform other clients in languages with a nominal type system as well.

As for my preference and kotlin specifically, I think that presence of utilities like Omittable is positively comforting that we're not completely in the weeds :)

Bnyro:
> I'm not sure I follow. Are you driving code-generation by parsing a CREATE TABLE statement? I would recommend driving code-generation from JSON schemas. Right now TB, doesn't help with that and defers to other, idiomatic JSON schema to your language converters (like quicktype).
>
Yes, I didn't know there are JSON schemas we can use instead of SQL for auto-generating classes. Just tested using https://github.com/wuseal/JsonToKotlinClass/blob/master/doc/LIBRARY.md, works, nice! If we decide to use Omittable, users would need to annotate with Omittable by hand where needed, but that shouldn't be too much of an effort (and you obviously only have to annotate it if you want to set a value to null).

> I'll defer to you as to what is idiomatic. I'm happy to absorb if it provides a canonical way to handling null vs absent.
>

> Would that something we could address if we absorb the Omittable?
>
We can make it slightly simpler by using something like `myObject.toOmittable()`, but not really much simpler. Because Kotlin is (in contrast to JavaScript) typesafe, we can't just pass a myObject directly, we have to manually wrap it with the `Omittable` type first.

ignatz:
> We can make it slightly simpler by using something like myObject.toOmittable(), but not really much simpler. Because Kotlin is (in contrast to JavaScript) typesafe, we can't just pass a myObject directly, we have to manually wrap it with the Omittable type first.

Many languages have a notion of implicit conversions (arguably magic, not sad that kotlin doesn't).

> If we decide to use Omittable, users would need to annotate with Omittable by hand where needed, but that shouldn't be too much of an effort (and you obviously only have to annotate it if you want to set a value to null).

If I follow this conclusion, it would mean we keep the code as is, i.e. skip nulls by default. We're saying overriding with explicit null is the odd one out that should require annotating.

If we go this route, would you still argue to absorb something like Omittable to minimize the footprint, or would it be reasonable to point users at Omittable?

Bnyro:
> If I follow this conclusion, it would mean we keep the code as is, i.e. skip nulls by default. We're saying overriding with explicit null is the odd one out that should require annotating.
>
Yeah, that's my verdict too.

> If we go this route, would you still argue to absorb something like Omittable to minimize the footprint, or would it be reasonable to point users at Omittable?
>
That's a good question. It's probably more convenient for users if the Omittable type is part of the TB client library already, so they don't need to read into the Omittable library. But strictly speaking, it's not really the scope of the TB client to provide kotlinx serializers.

You're the one to maintain it in the long term, so I'll leave the decision up to you. I can open a PR for a minimal Omittable, or update the docs in the Kotlin client Readme PR to point to the Omittable library instead. As you prefer :)

ignatz:
... or update the docs in the Kotlin client Readme PR to point to the Omittable library instead. As you prefer :)

I think this would be my preference. Gives me the chance to learn, gives us a workable start from which to improve, and we can always add more to the library (stealing candy is always harder).

ignatz:
And thank you so much for this wonderful discussion :pray:

ignatz:
After this discussion kept me a bit rattled not remembering the details of each and every client implementation (too many, too long), I went over most of them to at least document the state of the world and workarounds in tests.

As suspected, all the nominally typed language with some notion of serializable interface where the serialization is pushed into the library have similar challenges, after all how would they know. Luckily I feel a bit more competent in some of them.

Just for the sake of the discussion, let me point to some:

* dotnet: https://github.com/trailbaseio/trailbase/blob/dev/client/dotnet/test/ClientTest.cs#L26 has a bit more flexible annotations but basically you end up with types per use-case unless you fall back to explicit json objects.
* swift and kotlin have identical limitations/behavior, they currently require you to fall back to explicit json objects to ensure serialization:
  * https://github.com/trailbaseio/trailbase/blob/dev/client/swift/trailbase/Tests/TrailBaseTests/TrailBaseTests.swift#L263
  * https://github.com/trailbaseio/trailbase/blob/dev/client/kotlin/lib/src/jvmTest/kotlin/LibraryTest.kt#L378

I understand you're looking for something more type-safe and I'm on board. Custom serializers on `Omittable`-like wrapper types will probably work everywhere. But I'm somewhat comforted that no client makes it currently categorically impossible :hide:

Bnyro:
I quickly wanted to test the Omittable library before putting it into the docs, and I have bad news - it breaks when using `encodeDefaults = true` (even if the field doesn't have a default value).

E.g.
```kotlin
@Serializable
data class Foo(val nullableField: Omittable<Int?>)

Json.encodeToString(Foo(Omittable.absent()))
// yields {} as expected

Json {
  encodeDefaults = true
}.encodeToString(Foo(Omittable.absent()))
// yields {"nullableField":}, which is invalid JSON...
```

I don't get why it yields invalid JSON then, that seems to be some odd thing in Kotlinx serialization. At least I can't figure out how to fix the Omittable library, so implementing it ourselves wouldn't help. We're not the only ones with that issue, e.g. https://stackoverflow.com/a/72995768 says so too.

So unfortunately, I think that we can't take this path either unless we set `encodeDefaults` to false again and `explicitNulls` to true. Then people need to keep their database schemes in sync with the client models, but it is possible to use `Omittable` for specifying if data should be sent or not. So updating to null values is possible.

In my eyes it could even make sense to let the library user bring their own `jsonSerializer`, i.e. make it a public variable so that it's possible to switch it out depending on the use case.

ignatz:
You can technically make it work by combining omittable and annotations:

https://github.com/trailbaseio/trailbase/blob/omittable/client/kotlin/lib/src/jvmTest/kotlin/LibraryTest.kt#L492

Bnyro:
> You can technically make it work by combining omittable and annotations:
> 
> https://github.com/trailbaseio/trailbase/blob/omittable/client/kotlin/lib/src/jvmTest/kotlin/LibraryTest.kt#L492
>
True. That's not really convenient to use, but works.

ignatz:
It is verbose but my intuition is to keep `encodeDefaults = true`. Maybe it's my ignorance, but I would find it a very surprising kotlin-only feature that:

```kotlin
api.create(x);
```

may fail because there's a kotlin-level default erasing values for columns that are actually present.

Or that

```kotlin
val id = api.create(x);
val y = api.read(id);

assertEquals(x, y);
```

would trigger even if there were no writes in between because code and DB have mismatching defaults.

Bnyro:
> It is verbose but my intuition is to keep `encodeDefaults = true`. Maybe it's my ignorance, but I would find it a very surprising kotlin-only feature that:
> 
> ```kotlin
> api.create(x);
> ```
> 
> may fail because there's a kotlin-level default erasing values for columns that are actually present.
> 
> Or that
> 
> ```kotlin
> val id = api.create(x);
> val y = api.read(id);
> 
> assertEquals(x, y);
> ```
> 
> would trigger even if there were no writes in between because code and DB have mismatching defaults.
>
True, very good point. I'll update my PR with some more docs about it then and we keep the current behavior :+1:

Thanks a lot for discussing this topic, has been a pleasure <3

PR #288, opened 2026-09-12
Pr/native apple signin

ignatz:


crates/core/src/auth/api/apple_native.rs
```diff
@@ -0,0 +1,404 @@
...
+  pub auth_token: String,
+  pub refresh_token: String,
+  pub csrf_token: String,
+}
+
+/// Logs users in with a native Sign in with Apple identity token.
+#[utoipa::path(
+  post,
+  // QUESTION: Should this really be `/oauth/...`? What's the best practice here?
+  path = "/oauth/apple/native",
```
ignatz: For reference, it looks like authentic and keycloak are using:

* https://<keycloak-url>/realms/<realm>/protocol/openid-connect/token
* https://<authentik-domain>/source/oauth/callback/apple/

i.e. their standard token-exchange endpoints

crates/core/src/auth/oauth/providers/apple.rs
```diff
@@ -133,6 +53,12 @@ impl OAuthProvider for AppleOAuthProvider {
     return Self::NAME;
   }
 
+  fn auth_type(&self) -> oauth2::AuthType {
```
ignatz, line 56: IIUC, this is a bug in the current OAuth impl, where we're piggybacking a fix

crates/core/src/auth/apple.rs
```diff
@@ -0,0 +1,372 @@
+use serde::Deserialize;
+use sha2::{Digest, Sha256};
+
+use crate::auth::AuthError;
+
+#[allow(unused)]
+#[derive(Debug, Deserialize)]
+struct ApplePublicKey {
```
ignatz: There's nothing apple-specific about this, this is just a https://www.rfc-editor.org/info/rfc7517/#section-4

crates/core/src/auth/apple.rs
```diff
@@ -0,0 +1,372 @@
...
+    ));
+}
+
+/// Splits a compact JWT into its three base64url-decoded parts.
+fn split_jwt(id_token: &str) -> Option<(Vec<u8>, Vec<u8>, Vec<u8>)> {
+  use base64::Engine as _;
+  use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+
+  let mut parts = id_token.split('.');
+  let decode = |part: &str| URL_SAFE_NO_PAD.decode(part).ok();
```
ignatz: Just inline and use prelude.

crates/core/src/auth/apple.rs
```diff
@@ -0,0 +1,372 @@
...
+  let signature = decode(parts.next()?)?;
+  if parts.next().is_some() {
+    return None;
+  }
+  return Some((header, payload, signature));
+}
+
+/// Verifies signature and claims (issuer, audience, expiry) of an Apple
+/// identity token against the given public keys, selecting the key by `kid`.
+pub(crate) fn decode_id_token_with_keys(
```
ignatz: We could probably have a public fetch+decode, since both users use it as fetch plus subsequent decode. :shrug:

crates/core/src/auth/api/apple_native.rs
ignatz: We could also consider shoehorning this into oauth/callback since the code is pretty duplicative :/

yurvon-screamo:
@ignatz Hi! 

This looks great, thanks for taking it further - the split reads much cleaner now.

One small thing I noticed in the JWK cache (fetch_apple_public_keys): the shared future is stored in the cache before it resolves, so a failed fetch gets cached for the full 15-minute TTL - a transient network blip would fail all native logins for 15 minutes even after connectivity is back. Maybe only cache on success, or drop the entry on Err?

ignatz:
> @ignatz Hi!
> 
> This looks great, thanks for taking it further - the split reads much cleaner now.
> 
> One small thing I noticed in the JWK cache (fetch_apple_public_keys): the shared future is stored in the cache before it resolves, so a failed fetch gets cached for the full 15-minute TTL - a transient network blip would fail all native logins for 15 minutes even after connectivity is back. Maybe only cache on success, or drop the entry on Err?

Thanks for the review, much appreciated. I've since added an invalidation and some tests for the cache. If done a few more changes and cleanups and merged everything into the dev branch. This should go into the next release. Thanks for all your work :pray:

PR #292, opened 2026-09-14
Add README documentation to kotlin client

Bnyro:
draft until we found a solution for https://github.com/trailbaseio/trailbase/pull/287#issuecomment-5669132707


client/kotlin/lib/build.gradle.kts
```diff
@@ -75,6 +75,6 @@ mavenPublishing {
                 email = "contact@trailbase.io"
             }
         }
-        scm { url = "https://github.com/trailbaseio/trailbase" }
+        scm { url = "https://github.com/trailbaseio/trailbase/tree/main/client/kotlin" }
```
Bnyro, line 78: There doesn't seem to be any documentation url field according to https://maven.apache.org/pom.html, so not sure where we should put the link to the README.

ignatz: I've been clicking around trying to see what other more popular libraries do. I certainly didn't find a way to reference a readme directly. It think what you're doing here is probably all we can do.

client/kotlin/README.md
```diff
@@ -0,0 +1,83 @@
...
+// get everyone
+val persons = people.list<Person>().records
+
+// create a new person
+val person = Person(name = "foo", age = null)
+val id = people.create(person)
+
+// update an existing person
+person.age = 42
+people.update(id, person)
```
ignatz: Here you can see why the JSON schemas for read, create and update are all different. The update here also updates the name again, which may be an issue if the name has changed in the meantime. Ideally this would be:

```kotlin
val id = people.create(Person(name = "foo", age = null))
people.update(id, Person(age = 42))
```

which the singular Person bindings do not allow

client/kotlin/README.md
```diff
@@ -1,77 +1,83 @@
...
 ```kotlin
 client.login(username, password)
 ```
 
 ## Accessing records
-The TrailBase client can automatically deserialize the record API's JSON responses to Kotlin object.
-To do so, the Kotlin classes must be annotated with `@Serializable` from [`kotlinx.serialization`](https://github.com/kotlin/kotlinx.serialization).
 
-For example, this would look like
+You can either work with `JsonObject`s or let the client automatically (de)serialize records from and to Kotlin object.
```
ignatz: objects

client/kotlin/README.md
ignatz: After some sleep, I did end up spelling it out. I think it's important to demonstrate best practices and relevant for the subsequent skew discussion

ignatz:
Thanks for the docs, very much appreciated :pray: . Sorry for being so anal and doctoring so much. Personally, I find writing "good" docs worth reading way more laborious then writing "good enough" code. This is likely also why the current docs (or lack thereof) are so lackluster. That said, with us iterating on each other, I'm pretty happy with the concise, useful discussion that really every user of any client impl should be aware off :heart: . We should probably put a flavor of this front and center into the broader docs.

Bnyro:
Thanks a lot for the follow-up, I agree that it's much easier to read now and get started with!

I haven't got any experience with writing docs because I'm only working on end user applications, but not on libraries, so the feedback is greatly appreciated :)

PR #293, opened 2026-09-17
Add built-in helpers for expanded types to Kotlin client

Bnyro:
Currently, you would need to use two different data types if you want to expand columns:
- When sending data, only the id must be sent: `{"parent": "<id>"}`
- When reading data, the response layout is different: `{"parent": {"id": "<id>", "data": {...}}}`

The `Expanded` type wrapper basically allows you to do both with a single type.
- When serializing, it only serializes the id as a plain JSON string.
- When deserializing, it deserializes id and data.

I personally find this very helpful (I wrote this for my own project and decided to try to upstream it) because I can handle everything with a single data type, which makes it much easier to keep all models up to date.

I also understand if you don't like this approach though, it's possible to create this wrapper type in your app without modifying the Trailbase Kotlin client.


ignatz:
> I also understand if you don't like this approach though, it's possible to create this wrapper type in your app without modifying the Trailbase Kotlin client.

I actually like this a lot: nifty way of dealing with conditional expansion. Also an interesting precedent for other languages.

I have only low-value nits to offer:

```kotlin
children.create(Child(parent = Expanded(parentId)))
```

may sound a bit definitive. I wonder if 

```kotlin
children.create(Child(parent = Expandable(parentId)))
```

would reflect the conditional nature more (and follow Omittable's naming practices). I could also see an explicit initializer like `Omittable.Present()`, a la:

```kotlin
children.create(Child(parent = Expandable.Id(parentId)))
```

Ignorant question, how hard would it be to pull the impl (which is pleasantly straightforward) into a separate file? I've broken up all the other clients but I've always postponed kotlin to another day, simply all my tooling (LSP, ...) is generally unhappy with kotlin

ignatz:
Since we're already at discussing computer science's hardest problem, I could also see this being called `Reference` and s/id/ref/, e.g.:

```kotlin
val ref = Reference.Ref(id);
```

:woman_shrugging:

Bnyro:
> I have only low-value nits to offer:
> 
> ```kotlin
> children.create(Child(parent = Expanded(parentId)))
> ```
> 
> may sound a bit definitive. I wonder if
> 
> ```kotlin
> children.create(Child(parent = Expandable(parentId)))
> ```
> 
> would reflect the conditional nature more (and follow Omittable's naming practices). I could also see an explicit initializer like `Omittable.Present()`, a la:
> 
> ```kotlin
> children.create(Child(parent = Expandable.Id(parentId)))
> ```
> 
Sounds like a good idea, I'll adapt the PR to use the Expandable.Id naming.
> Ignorant question, how hard would it be to pull the impl (which is pleasantly straightforward) into a separate file? I've broken up all the other clients but I've always postponed kotlin to another day, simply all my tooling (LSP, ...) is generally unhappy with kotlin
>
It should work without any issues, it's as simple as just moving things to a different file and letting the IDE automatically import the other file. I use IntelliJ IDEA which I'm not happy with either, but I couldn't find anything else that works well for Android/Kotlin development. I'll already split the `Expandable` into a new file, and I can send a PR once this here is merged. I'm just not really sure if that breaks backwards compatibility with people that already use the library because the import paths might change.

> Since we're already at discussing computer science's hardest problem, I could also see this being called `Reference` and s/id/ref/, e.g.:
> 
> ```kotlin
> val ref = Reference.Ref(id);
> ```
> 
> 🤷‍♀️
>
I think that `Expandable` is more self-explaining because it matches the documentation (that talks about "expand") and makes it clear that the expanded thing can optionally be included.

`Reference` (at least to me) rather sounds like it points to some data, but doesn't actually have access to it really. (Although yeah, in C++ e.g. you do have access to referenced data, so idk).

ignatz:
> I think that Expandable is more self-explaining because it matches the documentation (that talks about "expand") and makes it clear that the expanded thing can optionally be included.

Happy either way. Keeping it close to the pre-existing terminology sounds wise.

> It should work without any issues, it's as simple as just moving things to a different file and letting the IDE automatically import the other file. I use IntelliJ IDEA which I'm not happy with either, but I couldn't find anything else that works well for Android/Kotlin development.

As someone who works across 10+ languages, I don't have strong feelings about IDEs but I have strong feelings about languages that want to choose for me :). Kotlin is a fine language but the tooling between LSP, gradle, JVM is one for the enterprises :woman_shrugging: 

> I'll already split the Expandable into a new file, and I can send a PR once this here is merged. I'm just not really sure if that breaks backwards compatibility with people that already use the library because the import paths might change.

If we're unsure about how import paths are handled, i.e. whether it's just the package or package.file, wouldn't `Expandable` be the perfect test-bed? At least after publishing we'll see where it ends up :upside_down_face:

ignatz:
Just saw the update, you already moved it to a separate file :pray:

client/kotlin/lib/src/commonMain/kotlin/Expandable.kt
```diff
@@ -0,0 +1,70 @@
...
+ */
+@ConsistentCopyVisibility
+@Serializable(with = ExpandableSerializer::class)
+data class Expandable<T> private constructor(val id: RecordId, val data: T? = null) {
+    @Suppress("FunctionName")
+    companion object {
+        fun <T> Id(id: RecordId) = Expandable<T>(id)
+        fun <T> WithData(id: RecordId, data: T?) = Expandable(id, data)
+    }
+}
```
Bnyro: I tried to use a sealed class instead, but then we would need to provide separate serializers for the `Expandable.Id` and `Expandable.WithData` types, which seems like an overkill to me.

Downside of the approach now is that we have to suppress the warning that function names should be lowercased :)

Bnyro:
> If we're unsure about how import paths are handled, i.e. whether it's just the package or package.file, wouldn't Expandable be the perfect test-bed? At least after publishing we'll see where it ends up 🙃
>
I'm 99% sure that if we keep everything in the same directory, nothing will break. At least the `Expandable` now has the same import path as the other classes in `Library.kt`.

client/kotlin/lib/src/commonMain/kotlin/Expandable.kt
ignatz: > Downside of the approach now is that we have to suppress the warning that function names should be lowercased :)

I don't have strong feelings. I guess we upper-cased it because of `Omittable.Present`. If lower-case is more idiomatic :yolo:

ignatz: FWIW, Record.(uuid|string|int|parse) are all lower-case

Bnyro: Even better, then we'll keep it lowercase :)

client/kotlin/local.properties
```diff
@@ -0,0 +1,8 @@
+## This file must *NOT* be checked into Version Control Systems,
```
ignatz: :)

Bnyro: Oops, I manually removed it from the added files for all previous commits, but forgot it this time.

I allowed myself to add it to `.gitignore` now :)

ignatz:
Did some more anal word-smithing. Implementation untouched. Looks great. Feel free to iterate and give me shit, otherwise I'll merge it

client/kotlin/lib/src/commonMain/kotlin/Expandable.kt
```diff
@@ -0,0 +1,73 @@
...
+            var id: String? = null
+            var data: T? = null
+            while (true) {
+                when (val index = decodeElementIndex(descriptor)) {
+                    0 -> id = decodeStringElement(descriptor, index)
+                    1 -> data = decodeSerializableElement(descriptor, index, serializer)
+                    CompositeDecoder.DECODE_DONE -> break
+                }
+            }
+            Expandable.withData(RecordId.parse(id!!), data)
```
ignatz: My only question was if it may be worth, to throw a more explicit exception 🤷‍♀️

Bnyro: The non-null assertion only fails if both these are true:
1. The returned object is a valid JSON dictionary/object (otherwise `decodeStructure` would already crash)
2. The object doesn't have an ID key

As far as I know, 1. is only satisfied if the user properly passes `expand=myColumnName`. So if the user forgets that, we should maybe catch the exception from `decodeStructure` and say "maybe you forgot to expand?".

For the 2. condition, I don't think it's reachable: All tables require to have a primary / ID key, so if the TB API returns expanded data, the `id` field is always set, right?

Bnyro: > As far as I know, 1. is only satisfied if the user properly passes expand=myColumnName. So if the user forgets that, we should maybe catch the exception from decodeStructure and say "maybe you forgot to expand?".
>
I've added error handling for that and a test case, I think that can indeed be very helpful for debugging.

Good idea to look into error handling here, I always forget about that :)

Bnyro:
> Did some more anal word-smithing. Implementation untouched. Looks great. Feel free to iterate and give me shit, otherwise I'll merge it
>
Thanks! I've squashed your changes into mine and fixed some other small typos (where we still talked about `Expanded` instead of `Expandable`).

ignatz:
Very nifty utility. Much appreciated :pray:

client/kotlin/lib/src/commonMain/kotlin/Expandable.kt
```diff
@@ -0,0 +1,81 @@
...
+                        0 -> id = decodeStringElement(descriptor, index)
+                        1 -> data = decodeSerializableElement(descriptor, index, serializer)
+                        CompositeDecoder.DECODE_DONE -> break
+                    }
+                }
+                Expandable.withData(RecordId.parse(id!!), data)
+            }
+        } catch (e: SerializationException) {
+            throw SerializationException(
+                """Failed to parse Expandable. Have you set `expand=listOf("<column_name>")` in the `read` or `list` call to expand the column?""",
```
ignatz, line 77: The second half may be a bit misleading, since parsing should work even when you omit expand. I guess this would rather be the issue if the column isn't expandable, i.e. not a FK or configured as expandable.

ignatz: All good, I might it a bit more specific to the missing id 🙏

ignatz:
Just release 0.7.6. Thanks again :pray:

client/kotlin/lib/src/commonMain/kotlin/Expandable.kt
Bnyro: Oh true! I didn't test that so i assumed that it's only wrapped into a dict with an `id` if `expand` is sent from the client, but in fact that's always done as soon as the API has the column marked as expandable.

Bnyro:
> Just release 0.7.6. Thanks again 🙏
>
Awesome, then I can already start using this in my app :)

client/kotlin/lib/src/commonMain/kotlin/Expandable.kt
ignatz: Coincidence? 😉

